Issue #96, opened 2022-11-05
Inform User of Device APK Architecture


S7venLights:
Just a nice to have, idea. 
Maybe some people who are new to tech and privacy won't know what APK to pick.
ARM, x64, x86, etc.
If there's a plugin that can detect which the phone supports, maybe there could be a message on the top of the pick APK dialogue that shows what architecture your device supports. 
Could read:
Your device supports: x, y and z

There are device info apps on F-droid that can pull this info somehow, but I don't know of a way to check this in android settings.

ImranR98:
This plugin should work: https://pub.dev/packages/system_info2/example

ImranR98:
https://github.com/ImranR98/Obtainium/releases/tag/v0.6.8-beta

PR #100, opened 2022-11-06
Add arch info

Marusiella:
Hi, I added architecture info in pop up dialog.
#96

### DIFF
--- a/lib/providers/apps_provider.dart
+++ b/lib/providers/apps_provider.dart
@@ -20,6 +20,7 @@ import 'package:path_provider/path_provider.dart';
 import 'package:flutter_fgbg/flutter_fgbg.dart';
 import 'package:obtainium/providers/source_provider.dart';
 import 'package:http/http.dart';
+import 'package:system_info2/system_info2.dart';
 
 class AppInMemory {
   late App app;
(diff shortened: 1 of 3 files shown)


ImranR98:
Thanks for contributing! I didn't realize this earlier, but we are already using a plugin this provides this info, so there's no need to add a new dependency. Could you switch to using `device_info_plus` (already added) and remove the new one? We get the supported device architectures with `(await DeviceInfoPlugin().androidInfo).supportedAbis`

lib/providers/apps_provider.dart
```diff
@@ -605,6 +628,9 @@ class _APKPickerState extends State<APKPicker> {
       scrollable: true,
       title: const Text('Pick an APK'),
       content: Column(children: [
+        Text(
+          'Your device ${widget.arch == "x86" ? "support" : "supports"}: ${widget.arch ?? 'Unknown'}',
+        ),
         Text('${widget.app.name} has more than one package:'),
```
ImranR98, line 634: I think we should put the architecture note below this text instead of above it. The architecture note should also have some visual separation - maybe put it in italics or make the font size smaller or both, whatever looks better to you.

ImranR98 (changes requested):
Looks good, just left a few comments for UI related changes.

ImranR98:
Also please do a `git pull` before pushing your changes to make sure you have the latest commits.

Marusiella:
I hope now it will be okay. Thank you for your advice, this is my first collaboration :).

Marusiella (commented):
fixed file

ImranR98 (approved):
Looks good, thanks!

PR #214, opened 2023-01-03
Added a cool splash screen, works in both light mode and dark mode and system

ToyZ-95:


ImranR98:
I appreciate the effort, but I don't think this should be added. Android already provides a simple splash screen that only lasts as long as needed. Ideally the user should be in the App instantly, and this PR adds an extra 2 seconds (plus another dependency).

Issue #219, opened 2023-01-07
Obtainium don't install updates


summoner001:
Hi!

Android version: Android 9
Obtainium version: 0.9.14-beta
New app installation permission: enabled
Storage permission: enabled

Clearing cache and storage don't solve the problem.

When Obtainiun notify me new updates are available for update. I tap to the down button. I don't leave the app when I'm waiting for download. When the download has complete obtainium don't install the update but the version tag changed to the new version number. But the app isn't updated.

Until now, the only time an update would not install for an application was when I left the application

The app: Mindustry Latest Bleeding Edge https://github.com/Anuken/MindustryBuilds/releases

ImranR98:
Most likely the App's `versionCode` has not changed between releases so Obtainium thinks you're trying to install the existing version. Please check if this is the case (if so, the developer would need to change it on their end). Will reopen if this is not the problem.

summoner001:
I checked the app version number in the phone options >> apps >> mindustry menu. The version number has older then the new version. And in the game menu has the older version number. Only the obtainium shows that the app it is updated to the new version.

When after obtainium download the apps the installer don't run. Only obtainium show the new version number.

Or how can I view the correct version number?

ImranR98:
Unfortunately Android only shows the version number, not version code. GrapheneOS does show it so I'll check and let you know. Though it is probably the case.

ImranR98:
Just checked - not a build number issue after all.

ImranR98:
Was looking at the wrong line. It is a version code issue. All versions of Mindustry show up as having a `versionCode` of `1`:
![Screenshot_20230107-163200-01](https://user-images.githubusercontent.com/30463115/211171096-1067a015-2a3d-4683-abf6-8b5852357721.jpeg)

This is something the app developer needs to fix, since the [versionCode](https://developer.android.com/studio/publish/versioning#versioningsettings) is supposed to be different for every version of an app.

That said, a few other apps have had this problem too so it may be a good idea to add a workaround for it in Obtainium.

ImranR98:
Fixed: https://github.com/ImranR98/Obtainium/releases/tag/v0.9.15-beta

Thanks for reporting this, found 1 other bug while fixing it too.

summoner001:
Yeah it is working for now. But the installer screen a bit laggy and freezing. After restart my phone the mindustry has updated. I will try to clearing obtainium cache and storage and I'll see if the installer freezing bug is still there.

Thank you for the fast coding!

Issue #149, opened 2022-12-14
Obtainium prompts for update but doesn't install the apk after downloading


Aayu1254:
**Apk installation skip because of unusual version number**
Obtainium shows update available when there's a new update and also downloads the update when i press update, but after downloading it just skip the installation and marks the app as updated. The app releases new updates as "latest" and not a version number in github.

**To Reproduce**
1. Add the app Xtra by crackdeded (Source: https://github.com/crackededed/Xtra)
2. Update the app through Obtainium

**Smartphone:**
 - Device: Realme X3 (RMX2081)
 - OS: Android 10
 - Obtainium Version: 0.8.9

ImranR98:
It looks like the developer doesn't increment the App's `versionCode` between releases. So as far as Obtainium is concerned, the downloaded APK doesn't need to be installed as it's identical to the old one. You're supposed to increment the `versionCode` for each version, so this is something the dev would need to fix on their end.

Also I wouldn't depend on reliable update notifications for this App since Obtainium checks the version by looking at the release tag and comparing it to the previous one. Would't work here as the dev uses "latest" as a tag. Again, tags are meant to identify a specific commit so reusing the same one doesn't make sense to me, and isn't something Obtainium can do much about.

PR #220, opened 2023-01-07
Same-versionCode Updates no Longer Skipped + 1 Other Bugfix

ImranR98:
- Obtainium would skip installing APKs that had the same [`versionCode`](https://developer.android.com/studio/publish/versioning#versioningsettings) since this number should be different for each new build of an App.
    - However, there are enough Apps that don't do this (#149, #219) so Obtainium now installs updates even if the `versionCode` has not changed.
- The GitHub release title filter has also been improved so that it filters by `tag_name` instead of `title` for releases with empty titles (as it seems like GitHub automatically displays the tag as the title in such cases).

### DIFF
--- a/lib/app_sources/github.dart
+++ b/lib/app_sources/github.dart
@@ -141,10 +141,13 @@ class GitHub extends AppSource {
         if (!includePrereleases && releases[i]['prerelease'] == true) {
           continue;
         }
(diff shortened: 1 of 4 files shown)


S7venLights:
🙌

Issue #30, opened 2022-09-24
F-Droid inclusion


Poussinou:
Hi,

Since this app is FLOSS itself, it could be a good idea to add it to F-Droid, so that people knows about it and use it.

What do you think?

ImranR98:
If it was on F-Droid, Obtainium would not be able to update itself (because F-Droid apps have their own signature). That's fine, but it means there would have to be a separate build of the which does not automatically include Obtainium as a tracked App out of the box (because then it would try to update itself from GitHub and fail). Right now that's hardcoded but can probably be changed to be optional based on a build option or something.

But is not a priority especially since part of the motivation for this app was to move away from F-Droid. Will leave the issue open but may not get around to it anytime soon.

sbstn87:
Perhaps a separate repo that can be added to F-Droid would be a solution for now?

rollingmoai:
> Perhaps a separate repo that can be added to F-Droid would be a solution for now?

Obtainium is now on the IzzyOnDroid repo: https://apt.izzysoft.de/fdroid/index/apk/dev.imranr.obtainium

ImranR98:
@rollingmoai does this require any maintenance when the app is updated? Or does it automatically grab everything from GitHub?

rollingmoai:
It automatically grabs everything from GitHub, so you don't have to worry about it.

ImranR98:
Great, thanks!

licaon-kter:
@ImranR98 
>  (because F-Droid apps have their own signature)

That's not a problem if the app is reproducible... but

> Obtainium would not be able to update itself 

F-Droid would like this to be opt-in :)

licaon-kter:
@momobobe the repro idea is _"the app is reproducible and F-Droid hosts the APK from the developer AFTER it verifies that reproducibilty"_

An in-app updater bypasses any verification, so users that put their trust in F-Droid get "lied" (strong word, bare with me) that all the app updates were verified.

FYI, almost every F-Droid cycle there's a repro app that fails verification.

So, to make things clear, an opt-in app updater is okay, as long as the user turns it on and gets informed that then the update comes straight from the developer.

inson1:
It there any upate?

linsui:
IIUC, we need to have the fdroid flavor build and uploaded somewhere.

ClockGen:
@ImranR98 can you please build and attach fdroid flavor builds as well to regular ones when you upload new releases?

[...]

PR #236, opened 2023-01-24
Add fdroid flavor

p1gp1g:
This PR add flavors to Obtainium: `github` for the main and `fdroid`. The FDroid flavor change the packageId and does not add itself to the app list.

That way:
- FDroid build has a different packageId
- FDroid build does not track itself on github.

This PR is useful only if the reproducible build is impossible (see https://f-droid.org/2023/01/15/towards-a-reproducible-fdroid.html). With this PR we can start the FDroid inclusion process (if you agree of course) and if the reproducible build doesn't work, this is a fallback solution :)

---

To run and build the main flavor:

```
flutter run --flavor github
flutter build --flavor github
```

For FDroid:
```
flutter run --flavor fdroid -t lib/main_fdroid.dart
flutter build --flavor fdroid -t lib/main_fdroid.dart
```

---

Related issue: https://github.com/ImranR98/Obtainium/issues/30

---
Edit: Change flavor name


ImranR98:
Hi, thanks for this PR. I'm not familiar with publishing on F-Droid but it [looks like](https://f-droid.org/docs/Submitting_to_F-Droid_Quick_Start_Guide/) we'd need to take some more steps to get that done. Would you be willing to set all that up as well? I've avoided the F-Droid issue so far as it doesn't seem worth the trouble to me, but we can go ahead with this if you're up for it.

ImranR98:
Could you change the name of the non-Fdroid build from `no` to `main`

p1gp1g:
Unfortunately, I can't, at least I have never found how to do it: gradle uses `main` under the hood already. It is still possible to use other names like `mainF` and `fdroidF`

ImranR98:
Makes sense. How about `standard`?

p1gp1g:
What do you prefer ? I am also thinking of `github` since it is the source intended to provide this build ?

ImranR98:
Yes, `github` is good

p1gp1g:
Done :)

notahomosapien:
Hi,
Please consider reading [this article](https://privsec.dev/posts/android/f-droid-security-issues/) before publishing on F-Droid.

p1gp1g:
> Hi, Please consider reading [this article](https://privsec.dev/posts/android/f-droid-security-issues/) before publishing on F-Droid.

1. Even if the app is published on f-droid, users don't have to install it from f-droid.
2. This blog post is very misleading. Most of the point from this blog post are outdated, others are also true for the playstore and/or the github releases or with an overestimated risk. Some points are still true.
3. This is not the right place to debate this :)

notahomosapien:
> This blog post is very misleading. Most of the point from this blog post are outdated, others are also true for the playstore and/or the github releases or with an overestimated risk. Some points are still true.

Have you actually read It?, Its not outdated and Obtanium is not going to be published on Google Play, therefore the only point that also applied to Google Play (The trusted party problem) is also not applicable. Github releases is different story, but at least Developers upload APKs signed by themselves and not by some third-party.

> This is not the right place to debate this :)

Okay, Thanks for informing me; Let's end this conversation?

linsui:
Any update?

ImranR98:
> Any update?

Sorry for ignoring this for so long, I will look at it soon. Once this PR is merged, I don't remember what the process is to get it published on f-droid so will need to refresh my memory.

linsui:
> I don't remember what the process is to get it published on f-droid so will need to refresh my memory.

No problem. Just ping me.

ImranR98:
Merging to dev first so I can test it

ImranR98:
Okay, the F-Droid flavour seems to build and run fine. Actually getting it up on F-Droid would be the next step.

PR #244, opened 2023-01-29
Add fastlane metadata for FDroid

p1gp1g:
Following discussion in #236 : this PR contains the metadata used for FDroid.

I open a PR on Fdroiddata with the package information and then post the link here. You welcome to review it ! (IDK if it will work as it is)


p1gp1g:
Here is the related PR on fdroiddata: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/12514

Maybe you would like to change a few things, like the AuthorName

ImranR98:
I recently updated the screenshots but you're using the old ones here.
Also could you remove the extra `u` at the end of the `AuthorName`.
For the build step, I think you still need to specify `--flavor fdroid` and we don't need `--target-platform` since we want builds for all architectures.
Aside from that, just confirming - did you try `fdroid build` to see if it works?

PR #253, opened 2023-02-02
Add flutter as a submodule for reproducible build

p1gp1g:
ATM, there is only one flutter app with the reproducible build on fdroid: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/12398

In order to get a reproducible build with flutter:
* flutter must be the exact same commit : that's why this PR set it as a git submodule (update the PATH afterward). [0]
* The full path to the project must be the same too. It is suggested to use /tmp/build : `ln -s $(pwd) /tmp/build ; cd /tmp/build`. Of course, if you want to use a github action for the build process, you won't even notice the path used :)

[0] Note: To fetch the current submodule code: `git submodule update` and to update it: `git submodule update --remote`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule ".flutter"]
+	path = .flutter
+	url = https://github.com/flutter/flutter/


p1gp1g:
Note: It might be a good idea to merge the fastlane PR and this one into a new branch to test the build process. You will have to give a signed build to check the reproducibility then :)

ImranR98:
Reproducible builds mean that the F-Droid version doesn't need a different App ID, right? Related to #236

p1gp1g:
> Reproducible builds mean that the F-Droid version doesn't need a different App ID, right? Related to #236

Yes indeed ! The PR #236 is a fallback solution if the reproducible build fails somehow, but I don't think it will be useful.

The next step is to test the reproducible build: we need a signed apk build from the fdroid-test branch, then we upload it to [fdroiddata PR](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/12514). We may need to send the commit hash too (most probably 1b48184)

PR #275, opened 2023-02-12
Add FA(Persian) language

mehdijahann:
Hi, I translated Obtainium to persian language ;)

Because the languages are right(persian is right to left) and left, after being applied, if there is a problem, I will fix it.


ImranR98:
Thanks, much appreciated!

PR #373, opened 2023-03-20
Rearranged Sources + Added (Non-Activated) WhatsApp Source (Website Currently Provides the Wrong Version)

ImranR98:


lib/providers/source_provider.dart
```diff
@@ -342,14 +343,15 @@ class SourceProvider {
...
+    FDroidRepo(),
     SourceForge(),
     APKMirror(),
-    FDroidRepo(),
-    SteamMobile(),
-    TelegramApp(),
+    Mullvad(),
+    Signal(),
     VLC(),
+    // WhatsApp(), // As of 2023-03-20 this is unusable as the version on the webpage is months out of date
```
kneejuicer, line 352: It seems that the APK on their website is now up to date. The version on the website (2.23.11.78) seems to be the same as the latest version on APKmirror.  It is even newer than the version on Google Play (for my device anyway - 2.23.11.77)

PR #465, opened 2023-04-13
Add an amoled black theme

### LINKED ISSUE #155: Add the amoled theme
It's a simple suggerition

Bnyro:
This PR adds a switch to enable a pure black (amoled) theme, works for both the Obtainium and the Material You theme.

closes #155

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -122,6 +122,7 @@
     "followSystem": "Follow System",
     "obtainium": "Obtainium",
     "materialYou": "Material You",
+    "useBlackTheme": "Use pure black dark theme",
     "appSortBy": "App Sort By",
     "authorName": "Author/Name",
     "nameAuthor": "Name/Author",
--- a/lib/pages/settings.dart
+++ b/lib/pages/settings.dart
@@ -224,6 +224,17 @@ class _SettingsPageState extends State<SettingsPage> {
                             ),
                             themeDropdown,
                             height16,
+                            Row(
+                              mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                              children: [
+                                Text(tr('useBlackTheme')),
+                                Switch(
+                                    value: settingsProvider.useBlackTheme,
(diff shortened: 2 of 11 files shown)


assets/translations/en.json
```diff
@@ -122,6 +122,7 @@
     "followSystem": "Follow System",
     "obtainium": "Obtainium",
     "materialYou": "Material You",
+    "amoledTheme": "Pure black theme",
```
ImranR98, line 125: Could you change this to "Use pure black dark theme" - this makes it more obvious that the toggle will only have an effect when the dark theme is on.

The key (`amoledTheme`) should also be closer to the value. Maybe `useBlackTheme`?

Also please add to all other language files (re-use the English version as a placeholder for others to update later).

lib/pages/settings.dart
```diff
@@ -226,6 +226,18 @@ class _SettingsPageState extends State<SettingsPage> {
                             height16,
                             colourDropdown,
                             height16,
+                            Row(
```
ImranR98, line 229: This should be directly below the `themeDropdown` as the two settings are related.

lib/pages/settings.dart
```diff
@@ -226,6 +226,18 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                              children: [
+                                Text(tr('amoledTheme')),
+                                Switch(
+                                    value: settingsProvider.amoledDarkTheme,
+                                    onChanged: (value) {
+                                      settingsProvider.amoledDarkTheme = value;
+                                    })
+                              ],
+                            ),
+                            height16,
```
ImranR98, line 240: `height16` won't be necessary anymore as the `colourDropdown` has enough space above it.

ImranR98 (changes requested):
Thanks for the PR! Just a few comments to review.

Bnyro:
Thanks for the review, I force pushed the suggested changes!

ImranR98 (approved):

PR #615, opened 2023-06-12
Fix out of memory error for large xapk's

1xFF:
This PR changes how Obtainium extracts zip files. Obtainium currently loads the entire zip file to memory, which will throw an out of memory error for large apps.

Testing has been hard for this given the time to download each 1-2GB app and possible further issues with how Obtainium handles OBB's/Multi-APK's in xapks. Although these apps no longer fail to install, I haven't yet had one successfully run, have it's expected app size, or have it's OBB's in android/obb/
So I'm putting this as a draft PR until it can be determined if my implementation is the issue or not.


1xFF:
Changing branch.

PR #616, opened 2023-06-12
Fix out of memory error for large xapk's

1xFF:
This PR changes how Obtainium extracts zip files. Obtainium currently loads the entire zip file to memory, which will throw an out of memory error for large apps.

Testing has been hard for this given the time to download each 1-2GB app and possible further issues with how Obtainium handles OBB's/Multi-APK's in xapks. Although these apps no longer fail to install, I haven't yet had one successfully run, have it's expected app size, or have it's OBB's in android/obb/
So I'm putting this as a draft PR until it can be determined if my implementation is the issue or not.

### DIFF
--- a/lib/providers/apps_provider.dart
+++ b/lib/providers/apps_provider.dart
@@ -27,7 +27,7 @@ import 'package:flutter_fgbg/flutter_fgbg.dart';
 import 'package:obtainium/providers/source_provider.dart';
 import 'package:http/http.dart';
 import 'package:android_intent_plus/android_intent.dart';
-import 'package:archive/archive.dart';
+import 'package:archive/archive_io.dart';
 
 class AppInMemory {
   late App app;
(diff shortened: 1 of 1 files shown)


1xFF:
Looking at cc3c4cc79f6ceb48f0cf5ee5bc9e9a5b161e3b5a it seems that OBB's are not supported, but there is code for installing multi-apks? That would explain issues with one app I had with just an APK and a OBB. It at least got to the loading screen, although oddly its installed app size was ~300MB when the base APK was only 80MB.

ImranR98:
Thanks for working on this. Yeah I haven't touched anything OBB related, not sure what could be causing the size increase.

1xFF:
@ImranR98 Do you know of an xapk app that works with Obtainium? It seems that if apkpure provides an xapk as a primary download it's because it has an OBB or large multi-apk's, otherwise it will be a regular apk.
For most apps you can download an xapk in the "all versions" page, but I can't get Obtainium to download that version using regex and it is unable to download xapks through an HTML source so I cant self-host the file.

ImranR98:
https://apkpure.com/microsoft-lens-pdf-scanner/com.microsoft.office.officelens/download

1xFF:
Thank you, I tried downloading that app through mainline obtainium but it says it's incompatible with my phone, presumably it's trying to use the x86 version.
I tested a small game with an OBB and confirmed the size difference happens on main branch too. The app started fine for both branches but obviously complained about the missing OBB. The issues seem to be unrelated to the PR so I'll mark it as ready for review.

PR #638, opened 2023-06-28
add OBB support

### LINKED ISSUE #622: Add support for obb files
Known issue, but couldn't find an open one.
Primarily related to xapks, but this also blocks full implementation of #343

1xFF:
Supports android versions below 11 and OS's with toggles (such as grapheneOS)

It's purported that apps with REQUEST_INSTALL_PACKAGES can access Android/obb [but that's either bugged or false](https://issuetracker.google.com/issues/254445476)

Android 11 and 12 has [a work around which is quite hacky](https://blog.esper.io/android-dessert-bites-28-file-manager-loophole-closed-73891524/), but if thats ok I can add support in another PR
Android API 34 has worked with MANAGE_EXTERNAL_STORAGE, but I'm not sure if this will still be the case in the final release of android 14 so I removed the code for it.

closes #622

### DIFF
--- a/lib/providers/apps_provider.dart
+++ b/lib/providers/apps_provider.dart
@@ -393,8 +393,11 @@ class AppsProvider with ChangeNotifier {
   void moveObbFile(File file, String appId) async {
     if(!file.path.toLowerCase().endsWith('.obb')) return;
 
-    // REQUEST_INSTALL_PACKAGES is required to access Android/obb
-    // But it seems impossible to check if obb access has been explicitly granted
+    // TODO: Does not support Android 11+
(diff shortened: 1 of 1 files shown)


ImranR98:
This is great. For Android 12+ the MANAGE_EXTERNAL_STORAGE is a good idea as long as it's optional (assuming Android 14 doesn't break anything again).

1xFF:
Thanks, I guess we'll need to wait till Android 14 comes out. MANAGE_EXTERNAL_STORAGE is a toggle but there doesn't seem to be a deny option. So whether to stop asking or not will have to be stored by the app...
I tried implementing the loophole for 11 and 12, but there's not really a good flutter plugin that would allow using SAF in the way needed.

Issue #427, opened 2023-04-05
Cannot remove app from app list


Zumux:
I am trying to remove Mullvad VPN from my app list. After clicking the remove button it appears to be removed until the app is closed and reopened and it reappears. From my limited testing this seems to be happening exclusively to the Mullvad VPN.

Anvil5465:
I've been having this issue for weeks. Usually I'm eventually able to delete the app from the list after a few tries, but at the moment there's an app in my list that I've deleted many times over the past couple weeks to no avail.

E0697:
This keeps happening for me.
I have to remove an app multiple times until it's finally gone for good.

E0697:
I use RealmeUI, Android 12.

ImranR98:
Is there any difference between deleting from the main list vs. the app detail page? Also do you usually uninstall at the same time as delete?

E0697:
I haven't tried it extensively but I think deleting from the app list works properly. (Will test and report back). Most of the time yes, I uninstall the app when removing.

1xFF:
I may have been able to replicate this bug. I got it to happen by uninstalling mullvad through its app settings and not obtainium, then quickly clicking the remove button before obtainium detects it's no longer installed.

That wont happen if you stay on the app page until the app icon disappears or if you uninstall the app through obtainium.

ImranR98:
Latest release contains a potential fix by @1xFF: https://github.com/ImranR98/Obtainium/releases/tag/v0.13.14-beta
Will close this issue in a week or so if no one reports it still happening.

1xFF:
@ImranR98 Probably safe to close

PR #647, opened 2023-07-03
fix race condition when checking updates

1xFF:
Found when looking into #427
Wont link this PR to close as I'm not sure if these bugs are the same.

While checkUpdate waits on getApp, removeApps can remove the appID from the list and delete the json file. After getApp returns, checkUpdate will rewrite the json file.
This isn't an perfect fix as these functions can run at the same time. The only other fixes I can think of would be either refactoring practically all of apps_provider, or somehow synchronously calling getApp which would cause a massive delay whenever an app page is opened.

### DIFF
--- a/lib/providers/apps_provider.dart
+++ b/lib/providers/apps_provider.dart
@@ -902,7 +902,7 @@ class AppsProvider with ChangeNotifier {
     if (currentApp.preferredApkIndex < newApp.apkUrls.length) {
       newApp.preferredApkIndex = currentApp.preferredApkIndex;
     }
-    await saveApps([newApp]);
+    if (apps.containsKey(appId)) await saveApps([newApp]);
     return newApp.latestVersion != currentApp.latestVersion ? newApp : null;
   }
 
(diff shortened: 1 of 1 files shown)


ImranR98:
No need for another if statement - `saveApps` already does this when the optional `onlyIfExists` parameter is set to `true` so we should use that instead.

1xFF:
I see, it looks like `onlyIfExists` is only used when updating the apps hashmap. Is this change how it should behave?

ImranR98:
You're right, I didn't catch that. New change looks good - just need to add `onlyIfExists` to the call to `saveApps`.

1xFF:
At the moment `onlyIfExists`'s default value is true, do you still want it to be explicitly passed as true?

ImranR98:
Right again. I clearly need sleep 🫠

1xFF:
Haha, I'm with ya. I almost pushed `onlyIfExists && this.apps.containsKey(app.id)`

PR #703, opened 2023-07-22
Fix #397 Edit Category Colors & Accessible colors using HSLuv

### LINKED ISSUE #397: [Enhancement] Customized categories color
Looks like there's no way to changing the color of categories. Any chance to have this function?

djm2k:
Love the app, I have switched from F-Droid to using Obtainium for all my FOSS android apps.
I tried organising by category but I kept getting different shades of green.

This PR makes category colors more accessible using a combination of the [HSLuv (Pastel)](https://pub.dev/documentation/hsluv/latest/hsluv/Hsluv-class.html) color space and incrementing by the [Golden Angle](https://en.wikipedia.org/wiki/Golden_angle).

I removed the randomness as it was causing some duplicate colors with the golden angle.

### DIFF
--- a/lib/components/generated_form.dart
+++ b/lib/components/generated_form.dart
@@ -370,6 +370,36 @@ class _GeneratedFormState extends State<GeneratedForm> {
                     }) ??
                     [const SizedBox.shrink()],
                 (values[widget.items[r][e].key]
+                as Map<String, MapEntry<int, bool>>?)
+                    ?.values
+                    .where((e) => e.value)
+                    .length == 1
+                    ? Padding(
+                    padding: const EdgeInsets.symmetric(horizontal: 4),
+                    child: IconButton(
+                      onPressed: () {
+                        setState(() {
+                          var temp = values[widget.items[r][e].key]
+                              as Map<String, MapEntry<int, bool>>;
+                          // get selected category str where bool is true
+                          final oldEntry = temp.entries.firstWhere((entry) => entry.value.value);
+                          // generate new color, ensure it is not the same
+                          int newColor = oldEntry.value.key;
+                          while(oldEntry.value.key == newColor) {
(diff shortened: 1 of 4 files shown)


djm2k:
I was attempting to add a button on the settings page to re-generate a category's color similar to the minus button deleting them currently, to close #397.

I couldn't figure out how the categories/colours are stored so I just use the workaround of removing and re-adding the category.

ImranR98:
> I couldn't figure out how the categories/colours are stored so I just use the workaround of removing and re-adding the category.

You can get `categories` from the `SettingsProvider`, and use `setCategories` to set them. The `categories` object is a `Map<String, int>` where the `String` is the category name and the `int` is the colour value.

ImranR98:
Other than that, great PR! Didn't realize I had that function twice.

djm2k:
Thanks for the tips, I managed to get it working!
Let me know of any issues, but I would love to see this released so I can switch off the debug app :)

ImranR98:
Nice 👀
Will be merged as soon as the `numColorsGenerated` comment is addressed.

djm2k:
@ImranR98 is that what you were after? Feel free to leave review comments for anything else
Thanks again!

ImranR98:
@djm2k no, there was nothing wrong with the link comment, I was talking about repeating colours. You should be able to see the review comments I left on the "Files Changed" tab: 
image

djm2k:
Thanks for the screenshot, I think those "pending" comments don't come through until you submit the review or something similar. I will add in your snippet now :)

djm2k:
@ImranR98 Should be good to go! let me know if there's anything else

ImranR98:
Whoops, you're right - I didn't actually submit the review comments. Looks good.

PR #934, opened 2023-09-23
Update pt.json

LucasTavaresA:


LucasTavaresA:
ok this fits all uses of appSource

Issue #1142, opened 2023-12-08
Apk cannot be installed since latest update from android. 


mollyuser:
**Prerequisites**

**Describe the bug**

**To Reproduce**

**Screenshots and Logs**

**Please complete the following information:**
 - Device: pixel 8
 - OS: GrapheneOS & Stock Android (tested on both)
 - Obtainium Version: all releases

**Additional context**

The apk cannot be installed since the last update from android, it is an android bug but it probably wont be fixed by them till february. I think it s best if u find a way to work around this bug because android users on latest release cannot install the apk anymore.

ImranR98:
Could you provide more details? I'm on the latest Graphene OS release and have had no issues installing Obtanium or apps from Obtanium.

mollyuser:
> Could you provide more details? I'm on the latest Graphene OS release and have had no issues installing Obtainium or apps from Obtainium.

I have reported this to GOS devs also, and they have confirmed the issue , it is also on stock os.QPR1 issue they sayd. I am not sure if only the pixel 8 series are affected. If obtainium is already installed everything works fine, the issue is only if u dont have obtainium and u download the apk and try to install it. It wont install. Are u on the 701 release for GOS?

mollyuser:
I got a message from GOS that they are going to fix the issue before google does, so the fix will be added in the next update. If u are on the latest GOS (701) the apk cannot install, i dont think u are on the latest release since it is confirmed that it s not device specific. StockOS will not fix it till february i think.

ImranR98:
Yes I'm on the 701 release. I have not tried reinstalling obtainium though, but updating it and installing other apps through it does still work.

Do you have any links/other info on this?

ImranR98:
Nvm, I see the chat on Matrix

ImranR98:
https://pastebin.com/8ea8HDsK

Well that sucks. I have nowhere near the Android knowledge needed to understand where the error comes from.

ImranR98:
GOS dev:

> it's fixed by the latest release that's coming out soon, please be patient
> if you were on stock OS, you'd be waiting for January or February

mollyuser:
GOS will fix it probably within a day or 2, but stockOS users will have to wait till january, probably february. Hope someone can help you to find a way to work around this bug.

mollyuser:
GOS devs are amazing, always on top of things.

skwde:
FYI, I had the same issue on Pixel 7a with GrapheneOS.

[...]

PR #1173, opened 2023-12-19
Resize adaptative icon to 108x108dp (#1142)

drizzt:
Android documentation [1] explicity specify that adaptive icons must be 108x108dp.
Having bigger Adaptive icons triggers a bug in Android 14 QPR1 that prevent to install apps from apk directly when the icons are bigger.

[1] https://developer.android.com/develop/ui/views/launch/icon_design_adaptive


ImranR98:
Nice, thanks!

PR #1184, opened 2023-12-24
Shizuku and Root

re7gog:
This is what you've all been waiting for


re7gog:
Tested on HyperOS(previously known as MIUI, where normal installation doesn't work) with KernelSU and Shizuku

ImranR98:
Thanks for this. I noticed that minsdk was increased but I guess it's okay? Given Android 6.0 is barely used anymore (https://www.businessofapps.com/data/android-statistics/), hopefully no one complains.

PR #1192, opened 2023-12-25
Fixes

re7gog:
Small important fix and Russian localization


re7gog:
Maybe one day I'll write obb file copier

assets/translations/bs.json
```diff
@@ -103,7 +103,6 @@
     "importErrors": "Uvezi greške",
     "importedXOfYApps": "{} od {} aplikacija uvezeno.",
     "followingURLsHadErrors": "Sljedeći URL-ovi su imali greške:",
-    "okay": "Dobro",
```
ImranR98, line 106: Good catch.

assets/translations/en.json
```diff
@@ -285,7 +284,8 @@
     "normal": "Normal",
     "shizuku": "Shizuku",
     "root": "Root",
-    "shizukuBinderNotFound": "Shizuku is not running",
+    "shizukuBinderNotFound": "Compatible Shizuku service wasn't found",
+    "installMethodsExplanation": "Alternative installation methods allow you to bypass the limitations of Android. This can, for example, solve problems when using firmware like HyperOS(MIUI).\nNormal or Shizuku is recommended.",
```
ImranR98, line 288: I don't think we need this. Someone who needs Shizuku or root probably already knows what that is. Besides it's very long - might make more sense to add this to the wiki instead of the app itself.

lib/pages/settings.dart
```diff
@@ -365,6 +363,28 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                            installMethodDropdown,
+                            height8,
+                            Text(
+                                tr('installMethodsExplanation'),
+                                style: Theme.of(context)
+                                    .textTheme
+                                    .labelSmall
+                            ),
+                            height8,
+                            GestureDetector(
```
ImranR98, line 375: I think the explanation and about link are unnecessary - they are only relevant to users who would likely not need them.

re7gog:
Now should be fine

ImranR98:
Why was this closed - something not working right?

re7gog:
I got confused in the commits, later I'll make another request
Also I want to try to make option to use system fonts

PR #1230, opened 2024-01-01
Fixes and system font

re7gog:


lib/main.dart
```diff
@@ -200,6 +201,16 @@ class _ObtainiumState extends State<Obtainium> {
                   context.locale.languageCode)) {
         settingsProvider.resetLocaleSafe(context);
       }
+      settingsProvider.addListener(() async {
```
ImranR98, line 204: We shouldn't need to store `appFont` in `settingsProvider` since it isn't used anywhere outside of this main build function. We also shouldn't need to use  `settingsProvider.addListener` since we're already watching the `settingsProvider` so the build function is re-run when any setting changes.

I tested the following and it works (replaces current lines 204-213):
```dart
      if (settingsProvider.tryUseSystemFont && fontToUse == "Metropolis") {
        NativeFeatures.tryLoadSystemFont().then((fontLoaded) {
          if (fontLoaded) {
            setState(() {
              fontToUse = "SystemFont";
            });
          }
        });
      } else if (!settingsProvider.tryUseSystemFont &&
          fontToUse != "Metropolis") {
        setState(() {
          fontToUse = "Metropolis";
        });
      }
```

Then we can have `String fontToUse = 'Metropolis';` initialized at the top in `_ObtainiumState`, and can remove the `appFont` getter/setter from `settingsProvider`.

assets/translations/en.json
```diff
@@ -287,7 +286,8 @@
     "normal": "Normal",
     "shizuku": "Shizuku",
     "root": "Root",
-    "shizukuBinderNotFound": "Shizuku is not running",
+    "shizukuBinderNotFound": "Сompatible Shizuku service wasn't found",
+    "tryUseSystemFont": "Try to use a system font",
```
ImranR98, line 290: I think this should say '_the_ system font` instead of '_a_ system font'.

Also in what situations will this not work? It worked on an Android 14 emulator without Shizuku. Does it depend on the Android version? If so we can hide the menu item on older versions.

lib/main.dart
```diff
@@ -236,13 +247,13 @@ class _ObtainiumState extends State<Obtainium> {
               colorScheme: settingsProvider.theme == ThemeSettings.dark
                   ? darkColorScheme
                   : lightColorScheme,
-              fontFamily: 'Metropolis'),
+              fontFamily: settingsProvider.appFont),
```
ImranR98, line 250: Replace with `fontToUse` as described above.

lib/main.dart
```diff
@@ -236,13 +247,13 @@ class _ObtainiumState extends State<Obtainium> {
...
           darkTheme: ThemeData(
               useMaterial3: true,
               colorScheme: settingsProvider.theme == ThemeSettings.light
                   ? lightColorScheme
                   : darkColorScheme,
-              fontFamily: 'Metropolis'),
+              fontFamily: settingsProvider.appFont),
```
ImranR98, line 256: Replace with `fontToUse` as described above.

lib/providers/settings_provider.dart
```diff
@@ -51,6 +51,24 @@ class SettingsProvider with ChangeNotifier {
...
+  String get appFont {
+    return prefs?.getString('appFont') ?? 'Metropolis';
+  }
+
+  set appFont(String appFont) {
+    prefs?.setString('appFont', appFont);
+    notifyListeners();
+  }
+
+  bool get tryUseSystemFont {
```
ImranR98, line 63: Can be removed, as described above.

lib/providers/settings_provider.dart
```diff
@@ -51,6 +51,24 @@ class SettingsProvider with ChangeNotifier {
...
+    return prefs?.getBool('tryUseSystemFont') ?? false;
+  }
+
+  set tryUseSystemFont(bool tryUseSystemFont) {
```
ImranR98, line 67: Can be removed, as described above.

ImranR98 (commented):
Great stuff, just have some minor suggestions.

lib/providers/settings_provider.dart
```diff
@@ -51,6 +51,24 @@ class SettingsProvider with ChangeNotifier {
     notifyListeners();
   }
 
+  String get appFont {
```
ImranR98, line 54: Can be removed, as described above.

lib/providers/settings_provider.dart
```diff
@@ -51,6 +51,24 @@ class SettingsProvider with ChangeNotifier {
...
+    return prefs?.getString('appFont') ?? 'Metropolis';
+  }
+
+  set appFont(String appFont) {
```
ImranR98, line 58: Can be removed, as described above.

re7gog:
Now Android 7.1.1-14 compatible (all app supported APIs) including MIUI, but error checking is left just in case.

Issue #1292, opened 2024-01-15
Support Direct APK Links


cautious-bread:
I've had a search for this in existing issues and also the HTML methods on the Wiki.

When attempting to add the Proton Drive APK directly, Obtainium appear to hang and then lock up, needing a force-close to allow it to open again.

1. Tap "Add App"
2. Paste "**https://proton.me/download/DriveAndroid/ProtonDrive-Android.apk**" into the "app source URL".
3. Tap the "add" button.
4. Watch the spinner for a few seconds before it stops, indicating the app has crashed.
5. Notice the app is not responsive then force-close it.

Trying the higher URLs e.g. **https://proton.me/download/DriveAndroid** or **https://proton.me/download** result in a "Forbidden" response.

No files of use under the /Andoid/data/ directory except for app json files and the Obtainiun cache.apk.

 - Device: Pixel 7a
 - OS: GrapheneOS
 - Obtainium Version: v0.15.11-beta

Nothing else to add - just hoping to get the Proton app APKs added directly.

ImranR98:
The HTML source is for HTML pages with APK links in them - it doesn't take direct APK links.

That said, we could add support for direct links easily now that #1101 has been resolved). I'll leave this issue open to track that feature, but you don't need it for your case. Adding https://proton.me/drive/download will work.

cautious-bread:
Thanks for that concise response @ImranR98 - received and understood.
Using the URL you suggested did the trick.

Cheers.

NicholasFlamy:
So what is the forbidden response? I assume it's because the website wants to prevent bots or something.

ImranR98:
https://github.com/ImranR98/Obtainium/releases/tag/v0.16.0-beta

Issue #1101, opened 2023-11-19
Suggestion for version detection in HTLM sources


hurried6158:
Currently, version detection for HMTL sources where there is no version string in the HTLM or the apk link is not supported. A case where such support would be nice is for protonapps.com as discussed in #970 and #606. I want to suggest an approach that might or might not be feasible in an Android app (my knowledge in android and app development for that matter is lacking).

I would suggest taking the checksum of something like the first 30 bytes (header of the zip/apk file) of the APK and store this. During a check for updates, only download 30 bytes of the APK, checksum and compare to the currently stored checksum. If they differ, one could conclude that there is a new version available.

Hopefully this approach is sensible.

/H

ImranR98:
Really good idea 👍

hurried6158:
I'm glad you enjoy the idea. I think the ideal way of doing this is with a HTTP range request. Not all servers hosting files support this, but many (or most?) do (this is the case with peotonapps). If they do, they should return HTTP response code 206 wheb the HTTP range request is processed. Since not all do, it might be worth having the version string support left as a fallback. And for scenarios with no support for HTTP range request, and no presense of a version string, it would be unsupported by Obtainium I suppose?

hurried6158:
(Sorry for accidentally pressing "Close with comment")

hurried6158:
So I did some experimenting by curling the Proton Calendar APK with a HTTP range requiest using ´curl´:

```bash
> curl -r 0-29 -O https://protonmail.com/download/CalendarAndroid/ProtonCalendar-Android.apk
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    30  100    30    0     0    110      0 --:--:-- --:--:-- --:--:--   109
> xxd -e ProtonCalendar-Android.apk
00000000: 04034b50 00000000 08210000 00000221  PK........!.!...
00000010: 00000000 00000000 00000000     0066  ............f.
```
Looking at the (APK) ZIP file header, unfortunately the relevant fields in the [header](https://en.wikipedia.org/wiki/ZIP_(file_format)#Local_file_header)  such as "File last modification date" and "CRC32 of uncompressed data" are all set to 0. This would probably mean that checksumming/hashing these 30 bytes won't really provide enough information that differs between two releases.

Perhaps one approach would be to checksum/hash something like the first 1KB of the APK. They should reasonably differ I suppose, but I can't say for sure.

[...]

PR #1303, opened 2024-01-20
Better partial APK hash as pseudo-version (related to #1101), Support Direct APK Links (#1292), Easier to understand version detection settings, Bugfix in automatic source selection for URL input, Various UI improvements and wording changes, Add invert option to apk filter regex , Filter out fdroid APKs from default Obtainium entry (for later)

ImranR98:
- Better partial APK hash as pseudo-version (related to #1101)
- Support Direct APK Links (#1292)
- Easier to understand version detection settings
- Bugfix in automatic source selection for URL input
- Various UI improvements and wording changes
- Add invert option to apk filter regex 
- Filter out fdroid APKs from default Obtainium entry (for later)


marek22k:
What does "Filter out fdroid APKs from default Obtainium entry (for later)" mean? Related https://github.com/ImranR98/Obtainium/issues/1308#issuecomment-1902174759 .

ImranR98:
GitHub releases now have F-Droid APKs in addition to regular ones (see the newest comments on #30). This means that anyone who tries to update Obtainium from within the app itself will need to manually pick between the regular and F-Droid variants. I added a filter in the default "Obtainium" entry that gets automatically added on new installs of the app so the user doesn't have to pick - F-Droid APKs will get filtered out automatically (on F-Droid variants, this entry is not added).

PR #1310, opened 2024-01-20
"Standardize" translation files + use machine-translations as placeholders for missing translations

ImranR98:


LilligantMatsuri:
Hi Imran, I'm a contributor to [Chinese translation](https://github.com/ImranR98/Obtainium/blob/main/assets/translations/zh.json).
Using machine translations as placeholders is a good idea, but there are strings that I prefer to leave untouched, e.g.,

https://github.com/ImranR98/Obtainium/blob/fc94d553eb21f5e39238c4ba7d96223bbe7c913e/assets/translations/en.json#L177-L178
https://github.com/ImranR98/Obtainium/blob/fc94d553eb21f5e39238c4ba7d96223bbe7c913e/assets/translations/en.json#L286

I wonder whether you replace all "untranslated" strings every time translation files are changed. If so, I suggest replacing only the newly added strings, and then I can easily find where to start.
Thank you for your awesome work! :blush:

ImranR98:
Sure, I plan to only replace newly added strings going forward

DwainZwerg (commented):
Hello Imran. In some cases you have replaced the same words in German as in English with a less adequate machine translation. I'll add a new German translation in a moment. Maybe you should let up on machine translations?

DwainZwerg:
I have a further suggestion: Use the translator deepl.com instead of Google Translate if the respective language is supported. It is more precise.

PR #1354, opened 2024-01-28
Add support for navbar color change on SDK 28

DanielKrasny:
Since some Android overlays (like Samsung One UI 1.0 with Android 9) are supporting dark theme system-wide, the minimum SDK for navigation bar control was changed from 29 to 28. I think this change should not restrict stock Android 9 users (or with some other overlay) from use.


ImranR98:
image

Unfortunately it does mess with the navbar on AOSP.

Issue #1381, opened 2024-02-08
Aurora Store/gitlab apps update version differs in normal and track-only modes


jayb-g:
**Prerequisites**
Install aurora store version older than currently latest version (older than 4.4.1 as of now)

**Describe the bug**
In track-only mode, it shows the update being 4.4.1 but as soon as the track-only toggle is turned off, it goes back to showing currently installed version as the latest one. This results into not being able to update to latest version. Not sure if this is Aurora/gitlab apps specific or generic issue. I've already tried removing cache but that didn't seem to have any effect.

**To Reproduce**

1. After installation as mentioned in prereqs above, edit the app config in obtanium and change it to track only mode.
2. Now it should show the latest update version (4.4.1 as of now) but you won't be able to install since its track only. 
3. Remove track only mode and it goes back to showing currently installed version as the latest version.

**Screenshots and Logs**
Track-only mode:
![](https://i.postimg.cc/05Y2Z3pT/Screenshot-20240208-213717-1.png)

After removing track-only mode:
![](https://i.postimg.cc/BZpfjG5n/Screenshot-20240208-213816-1.png)

**Please complete the following information:**
 - Device: [ Pixel 7 ]
 - OS: [ GrapheneOS ]
 - Obtainium Version: [1.0.2]

**Additional context**

ImranR98:
Track only mode has less strict requirements as Obtainium only looks for the version string, it doesn't need actual APKs. That's why track-only mode works fine. When you disable track-only mode, it tries and fails to find APKs - this is a known issue with Aurora (search for Aurora in the issues tab - there are several).

jayb-g:
@ImranR98 went through issues but it doesn't seem like unfixable since if obtanium is able to extract an older version, why not most recent version?(especially in case of aurora store?

After checking for differences in releases for 4.3.5 and subsequent ones, the difference is that 4.3.5 release had been attached but not released under _packages_ on gitlab; where's subsequent releases up to 4.4.1 are released under _packages_  and not as attachments. The download link for Aurora latest version is https://auroraoss.com/AuroraStore/Stable/AuroraStore_4.4.1.apk. Not sure why it would be a problem for obtanium to get the apk; even though I have provided the gitlab API Key.

![](https://i.postimg.cc/FFSwv1RB/Screenshot-20240209-102805-1.png)

![](https://i.postimg.cc/mrc8S615/Screenshot-20240209-102811-1.png)

[...]

Issue #1380, opened 2024-02-08
No update found at GitLab Release provider


mjdNjhNJ:
**Describe the bug**
[AuroraOSS/AuroraStore](https://gitlab.com/AuroraOSS/AuroraStore/-/releases/4.4.0) got an update today and Obtainium is not showing that the new version is available.

**To Reproduce**
1. Add [AuroraOSS/AuroraStore GitLab Release](https://gitlab.com/AuroraOSS/AuroraStore/-/releases/) URL to the Obtainium
2. Wait for a new version

**Please complete the following information:**
 - Device: Pixel 7a
 - OS: GrapheneOS (Android 14)
 - Obtainium Version: 1.0.2

**Additional context**
Can't update the app from version 4.3.5 to 4.4.0. Maybe the problem is that from AuroraStore 4.4.0 they are publishing the APK as "Packages" instead in attachements?

ImranR98:
@ThreeDeeJay is right - this likely a duplicate of the other Aurora issues.

DwainZwerg:
Hello @mjdNjhNJ ,
According to https://gitlab.com/AuroraOSS/AuroraStore/-/issues/1052#note_1778388859, the Aurora Store devs have fixed the problem. The download of AuroraStore should work normally again.

mjdNjhNJ:
@DwainZwerg I have updated the Aurora Store manually. Now Obtanium shows there is a new update available. However the update it's trying to install is Aurora Store with version of 4.3.5. The manually installed one is of version 4.4.1. Error message says: "Cannot install an older version of an App [Aurora Store]" which makes sense. I don't know if this is an Obtainium issue or the fact I mentioned while creating this issue.

DwainZwerg:
Hello @jayb-g ,
do you use GitLab-Token? If not, use it. If you are using it and Obtainium does not show you the update, you should delete the app from Obtainium and add it again. If it still doesn't work, it's an unknown behaviour, and I'm not sure if it's auroraoss.com's side or Obtainium. We are also in the process of adding a general feature to better retrieve updates from sites that are protected by Cloudflare (https://github.com/ImranR98/Obtainium/pull/1411).
#920 should actually have nothing to do with this, because it's about a problem with the F-Droid-API.
@ImranR98 @Disqu3-mirOir-qt for acknowledgement.

jayb-g:
@mjdNjhNJ its resolved with the token

Issue #697, opened 2023-07-17
Not finding latest version


Zumux:
Seems obtanium can't find the versions after 4.2.3 for Aurora Store

ImranR98:
In recent versions aurora store apks have been attached in a way that cannot be scraped without using the gitlab api. You should create a gitlab api key and enter it in the settings, you should see newer releases after that. See issue #3.

ImranR98:
Should be possible. Feel free to open another issue for it.

sereksim:
After adding the API key, the new version is found by Obtainium, but the download from https://auroraoss.com/AuroraStore/Stable/fails with the error 403: Forbidden.
-> It still is not possible to update Aurora with Obtainium :(

ImranR98:
The APK is hosted externally, not on GitLab, and the external site is blocking requests from Obtainium. See #854.

ImranR98:
This is due to a recent change on their end - it used to work.

DwainZwerg:
According to https://gitlab.com/AuroraOSS/AuroraStore/-/issues/1052#note_1778388859, the Aurora Store devs have fixed the problem. The download of AuroraStore should work normally again.

In addition, @akramer-zibra is working on a [PR](https://github.com/ImranR98/Obtainium/pull/1411) to avoid such problems with other websites that are protected by Cloudflare.

PR #1438, opened 2024-03-01
[discussion] Rework Gitlab .apk retrieval strategies

akramer-zibra:
`Obtainium 1.0.3` has two strategies to retrieve .apk information from Gitlab: This is either (a) by using the `Gitlab API` with a Private Access Token (PAT) or either (b) by using the `tags`-page of a repository. In some cases (e.g. Aurora Store) the `tags`-page does not have information about related .apk files, but the `Gitlab API` has (#1381, #1380, #697).

This patch will rework the strategy composition (a, b) and provide an interface to add other retrieval strategies, for example (c) scraping the `releases`-page of a repository. 

So the goal of this patch is to retrieve consistent .apk information with every retrieval strategy (API, tags-page, etc.) and in its consequence to make the `Gitlab API Token` superfluous in case of public repositories. Not every Obtainium user might have a Gitlab account to create a Private Access Token (PAT).

### DIFF
--- a/lib/app_sources/gitlab.dart
+++ b/lib/app_sources/gitlab.dart
@@ -20,28 +20,28 @@ abstract class RetrievalStrategy {
 class GitlabApiStrategy extends RetrievalStrategy {
 
   // Source context
-  AppSource source;
+  GitLab source;
 
   // Specific source parameters
   String standardUrl;
   Map<String, dynamic> additionalSettings;
 
-  // Specific strategy parameters
-  String privateAccessToken;
-
   /// Constructor with mandatory parameters
-  GitlabApiStrategy(this.source, this.standardUrl, this.additionalSettings, this.privateAccessToken);
+  GitlabApiStrategy(this.source, this.standardUrl, this.additionalSettings);
 
   /// Retrieves an iterable list of ApkDetails
   @override
   Future<Iterable<APKDetails>> retrieve() async {
+
+    // Load Private Access Token
+    String? PAT = await source.getPATIfAny(source.hostChanged ? additionalSettings : {});
 
     // Extract names from URL
     var names = GitHub().getAppNames(standardUrl);
 
     // Request "releases" data from API
     Response res = await source.sourceRequest(
-        'https://${source.hosts[0]}/api/v4/projects/${names.author}%2F${names.name}/releases?private_token=$privateAccessToken',
+        'https://${source.hosts[0]}/api/v4/projects/${names.author}%2F${names.name}/releases?private_token=$PAT',
         additionalSettings);
 
     // Check for HTTP errors
@@ -100,7 +100,7 @@ class GitlabApiStrategy extends RetrievalStrategy {
 class GitlabTagsPageStrategy extends RetrievalStrategy {
 
   // Source context
-  AppSource source;
+  GitLab source;
 
   // Specific source parameters
   String standardUrl;
@@ -291,7 +291,7 @@ class GitLab extends AppSource {
     if (PAT != null) {
 
       // Retrieve from Gitlab API
-      GitlabApiStrategy apiStrategy = GitlabApiStrategy(this, standardUrl, additionalSettings, PAT);
+      GitlabApiStrategy apiStrategy = GitlabApiStrategy(this, standardUrl, additionalSettings);
       apkDetailsList = await apiStrategy.retrieve();
 
     } else {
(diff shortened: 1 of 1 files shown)


akramer-zibra:
This patch comes also with the suggestion to refactor the .apk retrieval code into separate RetrievalStrategy classes and so decrease function complexity

akramer-zibra:
Hey @ImranR98 , is there a special reason Obtainium scrapes `tags`-page instead of `releases`-page in case of Gitlab repositories? The releases-pages may provide more relevant information as the tags-pages do 🤔

akramer-zibra:
> Hey, appreciate the effort

Oh, hey. And I appreciate your work and love for this project! Obtainium fills an essential part in my smartphone setup. This is why it would be a pleasure to give something back.

> What change is this referring to? From what I can see, you haven't modified either retrieval strategy - I only see that the existing code from GitLab.getLatestAPKDetails have been put into their own separate classes.

Yes. You are right. I did some reverse engineering to understand this code and have not yet added anything 😅. 

> I don't agree with the separation of GitLab specific code into a separate class outside of the GitLab class. Especially given that this new class has only one function (retrieve) and given that it needs you to pass an instance of GitLab into it anyways.

This is a common Design Pattern called `Strategy` (https://refactoring.guru/design-patterns/strategy), which is used to encapsulate "interchangeable algorithms" and so prevent functions from bloating. Sure, you are right. It seems a bit odd to give each `Strategy`-class a reference to the `Gitlab` instance, but this is due to the fact that `sourceRequest(...)` is generically defined within the `AppSource`-class and therefore is not accessible without a reference. Some further refactoring could resolve this, but I think this would go too far at this point.

> All pages have content that is loaded in with JavaScript and cannot therefore be scraped. 

You are absolutely right. I haven't seen this on my first glance. My bad! But nevertheless the JavaScript code e.g. on the `releases`-page uses Gitlab's public `GraphQL API` to retrieve the release information of a project. This API endpoint (https://gitlab.com/api/graphql) is public as well as Gitlab's `REST API`.

> We don't scrape the tags page, we download an RSS feed XML file from the tags page, then extract info from it (in fact it contains some info from the releases page including APK links).

Now I understand why Obtainium uses the RSS feed of the `tags`-page, but I do not understand why Obtainium uses the `REST API` only if a `Private Access Token` (PAT) is configured? The `REST API` does not need a `Private Access Token` to retrieve public data from public repositories. The Access Token is mandatory for private repositories etc. but not for public data 🤔. 

```dart
String? PAT = await getPATIfAny(hostChanged ? additionalSettings : {});

if (PAT != null) {
    GitlabRestApiStrategy apiStrategy = GitlabRestApiStrategy(this, standardUrl, additionalSettings);
    apkDetailsList = await apiStrategy.retrieve();
} else {
    GitlabTagsPageStrategy tagsPageStrategy = GitlabTagsPageStrategy(this, standardUrl, additionalSettings);
    apkDetailsList = await tagsPageStrategy.retrieve();
}
```

We already discussed in other comments that the `tags`-page and its RSS feed lack in some cases relevant .apk URLs (e.g. AuroraStore).

I do not see any reason why to use any other retrieval strategy than the official APIs, because both APIs will provide every public data available. A `Private Acces Token` is mandatory to access private repositories. Scraping other endpoints or pages or feeds does not lever this out.

For this reason, I see at least one option to improve the retrieval code:

Use either Gitlab's `REST API` or `GraphQL API` with an optional injection of the `Private Access Token` to access private data. 

What do you think @ImranR98 ?

ImranR98:
You're right - I remember thinking the API required a key even for public repos, not sure where I got that idea but it's not the case. Using the GraphQL API was also a great idea as a fallback but I guess it's not needed since the regular API does work w/o auth.

So we can just get rid of all tags RSS related code (remove the if-else) and only include the `?private_token=` query param if a token is defined.

ImranR98:
Also thanks for the link to https://refactoring.guru/design-patterns/strategy, seems interesting but probably not applicable here (I think the `AppSource` class as a whole could be considered as kind of like a Strategy class).

ImranR98:
Just checking - do you plan on updating this PR or should I go ahead and make the change?

akramer-zibra:
> Just checking - do you plan on updating this PR or should I go ahead and make the change?

Hey @ImranR98 . I can't get round to working on the code right now, so I am happy if you can implement what we found out 😀

akramer-zibra:
Hey @ImranR98 . I found some time to make the changes we discussion above. Therefore I created a new PR (#1459) and so close this one. As we also discussed above the refactorings in this approach seem out of scope.

Issue #1450, opened 2024-03-05
Aurora Store via Gitlab - the latest in Obtainium is 4.3.5 ?


acress1:
Hi all,

Aurora Store just published a week ago their 4.4.2 version (after 4.4.1 and 4.4.0), but in Obtainium, the latest available version is still 4.3.5.
I checked the settings and I believe everything is right.
Maybe there is something to check?

Thanks,
cress

DwainZwerg:
Hallo @acress1 ,
You are probably not using a GitLab token. Either you add a GitLab token or you wait for the next release of Obtanium, because then it will also be possible without a GitLab token (https://github.com/ImranR98/Obtainium/pull/1438#issuecomment-1975324096).

acress1:
Hi, yes sorry I forgot to enter my token..  Thanks a lot !

PR #1459, opened 2024-03-09
Improve gitlab .apk retrieval (#1450, #1381, #1380)

akramer-zibra:
Obtainium 1.0.4 has two strategies to retrieve .apk information from Gitlab: This is either (a) by using the Gitlab API with a Private Access Token (PAT) or either (b) by using the tags-page of a repository. In some cases (e.g. Aurora Store) the tags-page does not have information about related .apk files, but the Gitlab API has (#1450, #1381, #1380).

This patch removes the tags-page as an information provider and simplifies the retrieval logic. As it turned out: Gitlab's REST API works without `Private Access Token` authentication for public repositories. So the `Private Access Token` is optional in case authentication is required.

Previous discussion: #1438 

Tested with `Android 14` on an emulated AVD
- [x] _When_ I add AuroraStore with its Gitlab url _Then_ Obtainium shows the latest version correctly _And When_ I click install _Then_ Obtainium installs this version successfully
- [x] _When_ I add AppWarden with its Gitlab url _Then_ Obtainium shows the latest version correctly _And When_ i click install _Then_ Obtainium installs this version successfully

Notice: I haven't tested this patch with private repositories and/or private Gitlab instances yet.

### DIFF
--- a/lib/app_sources/gitlab.dart
+++ b/lib/app_sources/gitlab.dart
@@ -1,7 +1,6 @@
 import 'dart:convert';
 
 import 'package:flutter/material.dart';
-import 'package:html/parser.dart';
 import 'package:http/http.dart';
 import 'package:obtainium/app_sources/github.dart';
 import 'package:obtainium/custom_errors.dart';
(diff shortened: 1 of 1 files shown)


akramer-zibra:
Are there any texts in the UI which also should have been updated, too?

PR #1462, opened 2024-03-09
Removed outdated translations related to #1459

ImranR98:


DwainZwerg:
The search without the GitLab token is worse than with it, despite #1459, isn't it?

tatethegreat1st:
Jo

Issue #948, opened 2023-09-30
[Request] Can add it to the online translation platforms?（可以添加到在线翻译平台吗？）


ygxbnet:
I hope that the software translation of this project can be added to online translation platforms, such as [Crowdin](https://crowdin.com/) and [Weblate](https://weblate.org/).

This will greatly improve the efficiency of language translation, make the process more standardized, and reduce other maintenance work.

I am really happy to help you with this if I can, Thank you!

ImranR98:
Would there need to be changes to Obtainium's code to integrate with these services? Or would the translations made on these platforms end up being submitted as PRs to the existing `assets/translations/` folder?

ImranR98:
I see someone is already using Weblate for PRs: https://github.com/ImranR98/Obtainium/pull/713
Not sure if that is what you had in mind. I'm not familiar with Weblate.

ygxbnet:
> Would there need to be changes to Obtainium's code to integrate with these services? Or would the translations made on these platforms end up being submitted as PRs to the existing `assets/translations/` folder?

I'm at Crowdin. No need to change the code.

ImranR98:
Great, any PRs to add or update translations are welcome. I guess it wouldn't matter from the point of view of this repo where the PR comes from or what tools were used to make it.

ygxbnet:
But I hope we can have a unified platform so that all translations for other languages can be done on it, making collaboration easier.

ygxbnet:
And it can also allow more people to participate in the translation work.

ImranR98:
Makes sense. I'm guessing you'd run the Weblate or Crowdin page that others would then contribute to? Also, if any changes are made to the translations via other means (direct PRs) would those still be fine?

ygxbnet:
Yes, submitting PRs directly is still possible.

But because I don't have direct commit and branch creation permissions for the repository, it's challenging to set up a service that everyone can use. That's also the reason I submitted this issue.

ImranR98:
Okay, I might have to take a look at these tools to understand more about how they work.

ygxbnet:
Thanks!

Here are some simple differences:

**[Weblate](https://weblate.org/)** does seem relatively simple to use, but if you're using the official hosting, each user can only create one translation project. Additionally, since it's open source, you can also set up your own instance if needed.

[...]

PR #1543, opened 2024-04-15
fastlane: i18n ru for app description

yurtpage:
Please add the App Metadata to the Weblate.
Here is a sample of Conversations app metadata translation via Weblate https://translate.codeberg.org/projects/conversations/app-store-metadata-conversations/

Meanwhile I added one translation of the full_desscription.txt


ImranR98:
See #948

PR #1676, opened 2024-06-17
add Traditional Chinese translation

abc0922001:
add Traditional Chinese translation


ImranR98:
Hey @abc0922001, unfortunately this won't work in the app due to https://github.com/aissat/easy_localization/issues/372

PR #1740, opened 2024-07-21
Update README.md: Add OpenAPK

DwainZwerg:
Add the OpenSource-Repo “OpenAPK”.


ImranR98:
I'd rather not add this.
- We have an F-Droid button because some people prefer to use it (and since Obtainium uses reproducible builds, F-Droid is another party people can trust to check that my APKs are identical to the source code).
- We have an IzzyOnDroid button because they've been around for a while and are apparently trusted by many (not sure I would add the button now, but it's already there).

I haven't heard of OpenAPK and see no reason to recommend people get the app there instead of directly from GitHub.

DwainZwerg:
But that means sourcing directly from GitHub. It's a store that approves apps (like F-Droid, but still in its infancy) and then links to the latest GitHub release from the respective developer (ultimately, the website works like Obtainium, except that it curates like F-Droid).

PR #1749, opened 2024-07-25
Change the font (support for more languages)

VladWinner:
Native support for more languages: 
* Bosnian (bs)
* Russian (ru)
* Vietnamese (vi)

image

Source: https://fonts.google.com/specimen/Wix+Madefor+Display

PR #1793, opened 2024-08-16
Switch the font to Wix Madefor Text

FixItKit:
Obtainium's font changed to Wix Madefor Display in pull request https://github.com/ImranR98/Obtainium/pull/1749. Wix Madefor Text should be used for regular text (such as user interfaces) and Wix Madefor Display should be used for large text (such as branding) so Wix Madefor Text should be used instead of Wix Madefor Display.

https://fonts.google.com/specimen/Wix+Madefor+Text

(This pull request also capitalizes AMOLED in a code comment)


ImranR98:
Thanks for the PR but the display version looks closer to the original font used in Obtainium (Metropolis), so we'll stick with that. Also the about section for both fonts suggests the display/UI distinction depends on the font weight, not just the font used.

PR #1824, opened 2024-09-08
Update de.json

DwainZwerg:
see https://github.com/ImranR98/Obtainium/pull/1783#issuecomment-2310350237


DwainZwerg:
We had already agreed on the second commit. In general, Mr-Update had converted all constructions into infinitive constructions and made a small mistake there. The discussion afterwards centered on whether it was even necessary to convert the constructions that three? four? translators had worked on, for an infinitive construction. For reasons of consistency, however, everything should be an infinitive construction.

Issue #1752, opened 2024-07-28
make edits/cherry-picking of backups easier (for non-techies) and them thus more versatile


DJCrashdummy:
**Prerequisites**
sorry for rearranging the sections of the template a bit, but this isn't a thought out feature than rather an issue i encountered while real world usage...

**Additional context**
i'm using Obtainium for a while now and have recommended it to a friend of mine... as he bought a new phone and flashed it, he also wanted to start using Obtainium, but adding most of his apps one by one is a tedious task. so as a big chunk of the apps he uses, i use myself too, i wanted to give him a kick-start by sharing most of my apps added in Obtainium but not all, as he doesn't need all additional apps i'm only watching and - even more important - i don't want him to know exactly all apps i have installed on my phone. hence the normal export file was useless (unless i would start messing around with the json file manually). :unamused:

**Describe alternatives you've considered (if applicable)**
thankfully i'm a techy and curious person and already had a look at Obtainiums `Android/data/`-folder... there i found a folder neatly organized with one json file per app. :+1:
and simply copying a selection of these files into my friends `Android/data/`-folder worked like a charm. the chosen apps appeared in Obtainium and just had to be installed (if they weren't already).

**Describe the feature**
i'm asking, why are all json files merged into one for the export, even when they are internally used/stored one by one?
i suggest to also **export one json file per app**, that even non-techies can easily cherry-pick them before importing/sharing. then simply add an **additional file with the settings**, which e.g. would make me include them into the auto-backup just in case, as i know i can easily strip them from the backup afterwards (without having to mess around with a json file).
if the only argument against it is, that handling a single backup file is more convenient than a backup folder... my solution would be to pack all into a zip file or the like.
so we have both advantages combined: hassle-free handling of only one backup file which can easily be rearranged - perhaps even merged from different sources - by everyone and imported in one go.

ImranR98:
Instead of one file per app, it might be simpler to still use one file.

Apps page -> select multiple apps -> 3 dot menu ("more") -> "Share as export JSON"

The JSON could then be imported the same way as regular exports.

ImranR98:
https://github.com/ImranR98/Obtainium/releases/tag/v1.1.15

DJCrashdummy:
wow, that was quick! :flushed:

i already tried the new function... :expressionless: yes, it is nice to have. thereby **i** can choose which apps to export - or rather share - **manually**.

[...]

PR #1850, opened 2024-09-22
Format exported json

lagodimos:
Make the exported applications' file more readable in case it is needed to be opened with a text editor.


DJCrashdummy:
thank you very much @lagodimos! :+1:

this is a small but very solid step towards my suggestion mentioned some time ago at #1752.

PR #1915, opened 2024-10-20
Add Custom app ID

InfiniteCoder06:
Add Custom App ID generally used for Track Only Sources like ApkMirror
Without App ID
With Custom App ID

### DIFF
--- a/lib/pages/add_app.dart
+++ b/lib/pages/add_app.dart
@@ -540,11 +540,27 @@ class AddAppPageState extends State<AddAppPage> {
                   }),
             if (pickedSource != null)
               GeneratedForm(
-                  key: const Key('appId'),
+                  key: Key(
+                      '${pickedSource.runtimeType.toString()}-${pickedSource?.hostChanged.toString()}-${pickedSource?.hostIdenticalDespiteAnyChange.toString()}-appId'),
                   items: [
                     [
                       GeneratedFormTextField('appId',
-                          label: tr('appId'), required: false),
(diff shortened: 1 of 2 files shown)


ImranR98:
Not sure it's a good idea to let users change App IDs. They are meant to be unique and unchanging, and a lot of features have been written with this in mind (even app entry/config data is stored in a JSON file named after the app ID). Have you tested this thoroughly to make sure it doesn't break anything?

lib/pages/add_app.dart
```diff
@@ -538,6 +538,22 @@ class AddAppPageState extends State<AddAppPage> {
...
+                  items: [
+                    [
+                      GeneratedFormTextField('appId',
+                          label: tr('appId'), required: false),
+                    ]
+                  ],
+                  onValueChanges: (values, valid, isBuilding) {
+                    if (!isBuilding) {
+                      setState(() {
+                        additionalSettings['appId'] = values['appId'];
```
ImranR98, line 553: This value should be reset when the rest of the form is reset (when the picked source changes).

ImranR98:
Still not sure about this. Will need to test it more.

InfiniteCoder06:
Added Validation of App ID (Package Name) as well

> Still not sure about this. Will need to test it more.

ImranR98:
It shouldn't be possible to use this when an app is not added as track-only (should be hidden). Also could you describe what ways you have tested this

PR #1929, opened 2024-10-24
Update Polish translation

Daviteusz:
This is probably my last translation update, because this automatic "transkations" BOT is ruining everything.

I translated from scratch everything that the bug BOT had scrawled.


ImranR98:
Sorry, the auto translator picks up any string that's in English, except for items that have been added as exceptions. I'll fo through your PR and add anything you reverted back to English as an exception so the auto translator ignores it next time.

ImranR98:
@Daviteusz on closer inspection, the auto-translator only added a translation for `"placeholder"`, nothing else. All the remaining changes were just a re-ordering of entries to be consistent with all other translation files. Is there any reason you need a custom/different order of entries in `pl.json`?

Daviteusz:
I did not notice that the program in which I translate changed the sorting. I will correct it in the next PR. 

Regarding the auto translation bot, I got a little carried away 😄

Issue #1986, opened 2024-11-19
German Translation


PrivacyAndSecurity:
Hey,

i noticed a mini Translation thing in the Settings. It says: "Pure Black Dark Theme verwenden" could you change that to: "Rein schwarzen Hintergrund verwenden" ? Because that would be more accurate German Translation. Otherwise it is mixed. I would appreciate that a lot!

ImranR98:
https://github.com/ImranR98/Obtainium/releases/tag/v1.1.32

Issue #1973, opened 2024-11-14
"Forbidden" error when adding or refreshing APKMirror track-only apps


bossanovaorca:
**Prerequisites**

**Describe the bug**

APKMirror applications are track-only on Obtainium, but when refreshing or trying to install new apps on APKMirror, Obtainium returns "Forbidden". It doesn't seem to be related to my network so I'm not sure why Obtainium is having trouble with this site.

**To Reproduce**

1. Add an APKMirror application to Obtainium, for example https://www.apkmirror.com/apk/x-corp/twitter/ (it will be "track-only")
2. look for any "Forbidden" errors

**Please complete the following information:**
 - Device: Samsung Galaxy S23 Ultra
 - OS: Android 14
 - Obtainium Version: 1.1.30

I can provide any logs or screenshots if needed

leo-liar:
Can report the same behaviour since a few days ago (with a different app). Update check suddenly failed with "Forbidden". Removed the app from Obtainium but could not add it again because this action is also "Forbidden".

- Device: OnePlus 8T
- OS: Android 13
- Obtainium version: 1.1.30

ImranR98:
Yeah it seems like cloudflare protection has gone from being inconsistent to always on. May have to remove the apkmirror source if a solution is not found.

bossanovaorca:
oh ok I did see #1097 when searching for previous tickets but I wasn't aware that my current issue was Cloudflare related. I may have to look for other sources to migrate to then

DwainZwerg:
@archon810 Do you have an idea?

DwainZwerg:
Do APK Grabber and APK Updater also have the problem? Wenn nein, könnten wir deren source Code übernehmen.

archon810:
Try setting a proper user agent that references the app being used to query ("Obtainium" with the version would be a good start).

fahim-ahmed05:
Try with common browser user agent

Vinfall:
Pretty much a *intended behavior* if I have to guess.

In the past APKMirror also blocked apkeep by EFF, see `https://github.com/EFForg/apkeep/issues/46#issuecomment-1308244059` (quoted to avoid linking issue). But that was 2022, things might have changed.

Also, since APKMirror source in Obtainium is track-only, users need to visit APKMirror site for downloading, maybe they would (still) like to

> welcome and collaborate with apps that try to integrate with the site and direct users to our download pages, or even facilitate an API level collaboration

[...]

Issue #682, opened 2023-07-16
Updating from apkpure fails for certain apps (others work):


ArnoldSmith86:
![Screenshot_20230716-110838_Obtainium.png](https://github.com/ImranR98/Obtainium/assets/73538315/1981ba23-80e2-443f-9d7a-82f87ef100a4)

![Screenshot_20230716-110852_Obtainium.png](https://github.com/ImranR98/Obtainium/assets/73538315/58a0a8cd-a043-436e-8a41-d2e04fb81146)

What details do you need?

ImranR98:
I'm able to install this with no issue. Could you try installing it manually without Obtainium?

ImranR98:
Actually try it with this new release first: https://github.com/ImranR98/Obtainium/releases/tag/v0.13.18-beta

ArnoldSmith86:
With new release: ![Screenshot_20230717-124911_Obtainium.png](https://github.com/ImranR98/Obtainium/assets/73538315/c48420a7-7a60-4700-8340-9dfcaf7b7cd9)

ArnoldSmith86:
After uninstalling first: ![Screenshot_20230717-125110_Obtainium.png](https://github.com/ImranR98/Obtainium/assets/73538315/8c49b1a5-fbf7-483f-b8c3-b50550dd95b3)

ArnoldSmith86:
Installing using https://f-droid.org/en/packages/com.smartpack.packagemanager/ worked fine:

![Screenshot_20230717-130210_Package Manager.png](https://github.com/ImranR98/Obtainium/assets/73538315/d337a628-43de-4364-8968-8525a815f726)

Not sure how else to install XAPKs.

1xFF:
I wonder if that config apk is actually being installed by that app

ArnoldSmith86:
It lets me select what to install. I have no idea how all that xapk stuff works, though.

ArnoldSmith86:
Not sure if this helps but:

```
:/data/local/tmp # cp /storage/emulated/0/Android/data/dev.imranr.obtainium/cache/com.agilexrobotics-523288793.xapk-dir/*apk .
:/data/local/tmp # ls -l
total 71180
-rw-r----- 1 root root      34735574 2023-08-16 10:39 com.agilexrobotics.apk
-rw-r----- 1 root root         45465 2023-08-16 10:39 config.ar.apk
[...]

PR #1990, opened 2024-11-23
Improved XAPK Support (#682), Custom user-agent for APKMirror (as per feedback in #1973), Minor change to German translation (#1986), 

ImranR98:
- Improved XAPK Support (#682)
- Custom user-agent for APKMirror (as per feedback in #1973)
- Minor change to German translation (#1986)


lib/app_sources/apkmirror.dart
DwainZwerg, line 1: If I unverstood the discussion in the issue correctly, you not online wanted to edit the User-Agent to Obtainium/version number but also give the abbility to edit the User-Agent by ourself. In this commit I don't see anything related with the second part. In the newest release I also couldn't change the User-Agent.

PR #2090, opened 2025-01-26
Fix and improve German translation 

ColorfulRhino:
- [Restore two lines in de.json accidentally removed during merge conflict](https://github.com/ImranR98/Obtainium/commit/1e8fd3346970e2fb34ee509556eb5d3b21027d49)
- Increase consistency for certain terms (e.g. reverting to use "Nachverfolgung" in one instance)
- Address concerns for the translation of _author_ containing a colon by using a different wording (_publisher_). I was able to mostly solve the discussion about inclusive language by using an abbreviation, which is needed anyway to prevent the word from being truncated
- Minor wording improvements

(As discussed in #2059. This addresses some concerns voiced in that PR)
@DwainZwerg feel free to take a look.

This slightly changes the meaning of _author_ to _publisher_. In my opinion _publisher_ instead of _author_ better reflects the actual context in English as well. Why? While uisng e.g. GitHub as source, Obtainium will show the user or org from the repository. In case of single-dev apps this is often the actual author, but you may argue that for example Fossify is not considered the author of [Fossify Gallery](https://github.com/FossifyOrg/Gallery), but multiple contributors are. This is just a nitpick though and some may argue otherwise.
However, e.g. with IzzyOnDroid, Obtainium does not show the actual maintainer of the app, but rather only shows IzzyOnDroid in Obtainium's app overview. Izzy is certainly not the author of those apps, but the person publishing/uploading them to their repository.
Thus using the word _publisher_ instead of _author_ should be evaluated since it more broadly reflects what Obtainium currently wants to say with this term.
What's your @ImranR98 and community's opinion on this?
_(I do realize that this is just a nit and many likely won't ever care or notice this)_


ImranR98:
I think it makes sense to go with Publisher for the reasons you outlined. May make the same change in English later.

Issue #2013, opened 2024-12-09
Multi-device sync/recover capability


Fmstrat:
**Describe the feature**
If open to it, considering authoring a PR for full-sync and recover. This is similar to previous sync requests like backing up to cloud providers, but would handle the sync the same way applications like Steam or Google Play do.

UI additions:
- Toggle for sync
- Sync server URL selection
- Device name (defaults based on similar name pattern to Firefox)
- Sync credentials
- Additional dropdown on restore screen for "Install to match <device>/none
- In each app details, a list of "Installed on" devices

Backend additions:
- Current backup format in JSON is used, but 
  - Each installed app gets an Installed On array of devices
- When opening Obtainium and/or adding a new app, a sync occurs
- Adding apps is not allowed if sync fails
- When installing/uninstalling/removing, a date time flag to indicate action is added in an Actions array
- Removing an app keeps the app in the configuration, but hides it from the UI

Sync method:
- For each app, whichever Action has the latest date (device or server) is used
  - If server has newer action than device, and server says installed for the current device but device says it is not, an Action of ResolveUninstalled is added to the array and saved in both locations
    - This is ignored if "restore" is in progress
    - Potential v2 revision to this to allow for "install on another device"

Still debating on the sync server infrastructure. I could pretty easily make a docker-based service with a postgres DB, but users may prefer a selection of cloud providers like Drive and Nextcloud. The advantage of the former is you (or anyone) could run a multi-user sync service, which could be a pathway to revenue if you so desired (I would only submit a PR under the current license, though).

**Describe alternatives you've considered (if applicable)**

Current export/import

**Additional context**

I want to be able to use Obtainium across multiple devices fully de-googled. This would be a lot easier if Obtainium supported a centralized account structure.

Fmstrat:
I should.probably note, I have a good bit of experience in sync. I authored OwnNote for OwnCloud, a sync service for Plex, an Rclone system for Steam Deck retro backups, and many moons ago helped debug port issues woth Firefox Sync.

ImranR98:
I have closed similar issues in the past because it's not something I would undertake (usually issues I don't plan to work on do remain open for others to contribute, but I didn't think there would be any PRs for something this big). But if you're planning to contribute a PR, that would definitely be accepted after review. Personally I'd prefer the Docker-based self-contained option since it's less hassle (Nextcloud has always been very frustrating/buggy for me) but that's up to you of course. And yes, the license is not going to change.

[...]

PR #2149, opened 2025-02-26
Add multi-device synchronization support

Fmstrat:
This PR incorporates synchronization functionality with [SxncD](https://sxncd.com). SxncD is a self-hostable, FOSS synchronization system built for apps that currently use file import/export with a plugin already written for Obtainium. As you can see from the screenshots below, Apps each get their own API key per user, and each App has a card representing it. The selection of apps and authentication methods can be made when setting up the server.

The best way to think of it is SxncD is for synchronization what [https://ntfy.sh/](ntfy.sh) is for notifications.

Covers issue https://github.com/ImranR98/Obtainium/issues/2013.

<table cellpadding="10" cellspacing="0" border="0" width="100%">
    <tr>
        <td align="right">
            
        </td>
        <td align="left">
            
        </td>
    </tr>
</table>

I've been running this on multiple devices for a couple of weeks now and it's running smoothly. I'm labeling this PR for draft as I'm not sure how translations are handled for new content yet.

Overall, this PR:
- Updates a timestamp every time settings are changed or an App is updated/added/removed
- Synchronizes on startup or on manual refresh
- Adheres the the import/export setting of "Include settings" for the synchronization (app only or app + settings)
- Warns users that remote settings will overwrite local settings on first sync if the Device Name is not pre-existing (I.E. setting up a new device)


ImranR98:
Translations are handled in the way you've done - add English in all files until someone contributes the translations (although I usually use the [standardize.js](https://github.com/ImranR98/Obtainium/blob/main/assets/translations/standardize.js) script to do this automatically (ensures consistency) and automatically add AI translations from DeepL as placeholders. You can do this if you want to and have an API key, otherwise I'll do it later.

ImranR98:
What is this [core](https://github.com/ImranR98/Obtainium/pull/2149/files#diff-0d45f5fd462b8c70bffb10021ac1bcff3f58f29b1faf7568595095427d42812c) file?

Fmstrat:
> You can do this if you want to and have an API key, otherwise I'll do it later.

I don't have one, but I'm happy to get one together if it's a pain for you to do.

> What is this [core](https://github.com/ImranR98/Obtainium/pull/2149/files#diff-0d45f5fd462b8c70bffb10021ac1bcff3f58f29b1faf7568595095427d42812c) file?

Doh, that was a core dump from a failed run of the flutter emulator that must have been left over from the `docker` feature branch when I was setting that up. Removed. Luckily it's marked binary so it's fully removed from the history (since it's 1Meg in size) with the latest commit.

ImranR98:
> I don't have one, but I'm happy to get one together if it's a pain for you to do.

Nah it's no big deal
Thanks

ImranR98:
Can't seem to get this to work yet. I tried setting up SxncD but couldn't get OIDC auth to work due to https://github.com/Fmstrat/SxncD/issues/2. Then I tried GitHub auth and that worked on the webapp side - I signed in and got an API key for Obtainium.

When I try syncing from my phone, I get an error about an "Unexpected character" in "`<!DOCTYPE html...` which makes sense because we don't expect an HTML response. In the server logs, we see the problem: `PayloadTooLargeError: request entity too large`. I only have 52 apps and my JSON export is only 173 kB so I'm not sure why this is happening.

Then I tried using the Android emulator (only 10 apps on it). This time the payload was small enough, but then I got this error: `insert or update on table "app_device" violates foreign key constraint "app_device_app_id_fkey"`. Not sure what's causing this. My guess is that the emulator is being assigned a device ID of `1` which may have already been assigned to the phone due from the failed sync earlier? Haven't actually confirmed this.

Overall it seems like it needs more testing around edge cases and failure modes.

ImranR98:
Also, what happens if I have, for example, version `1.1` of some app `XYZ` installed on one device, but version `1.2` on another? If I'm reading this right, you're syncing the app install states, which I was concerned about in https://github.com/ImranR98/Obtainium/issues/2013#issuecomment-2528951728 - this would cause devices to show inaccurate install state info, especially for apps that do not have the "reconcile installed version..." toggle enabled. Please correct me if I'm wrong, maybe I missed something.

Fmstrat:
> Can't seem to get this to work yet. I tried setting up an SxncD server but couldn't get OIDC auth to work due to [Fmstrat/SxncD#2](https://github.com/Fmstrat/SxncD/issues/2). Then I tried GitHub auth and that worked on the webapp side - I signed in and got an API key for Obtainium.
> 

Interesting, I've used it with Keycloak and a custom Node server implementation and both worked for me. Happy to help debug config if you'd like.

> When I try syncing from my phone, I get an error about an "Unexpected character" in "`<!DOCTYPE html...` which makes sense because we don't expect an HTML response. In the server logs, we see the problem: `PayloadTooLargeError: request entity too large`. I only have 52 apps and my JSON export is only 173 kB so I'm not sure why this is happening.
> 

Odd, let me try adding a bunch of apps to mine and see what happens.

> Then I tried using the Android emulator (only 10 apps on it). This time the payload was small enough, but then I got this error: `insert or update on table "app_device" violates foreign key constraint "app_device_app_id_fkey"`. Not sure what's causing this. My guess is that the emulator is being assigned a device ID of `1` which may have already been assigned to the phone from the failed sync earlier? Haven't actually confirmed this. The IDs should probably be random UUIDs or a hash of the user ID + device name.
> 

I'm wondering if this has to do with the above error. I realize I haven't included each interaction in a transaction, so it's like the DB just got into a bad state (amateur hour on that one). Will take care of that next.

> Overall it seems like it needs more testing around edge cases and failure modes.

Oddly I tried a bunch of things to break it, and have a full e2e test running. I think in the above you may have triggered a cascading issue off the doctype item. In any event, I'll find a way to reproduce and make sure that's covered in the test cases. Thanks for helping out.

> Also, what happens if I have, for example, version `1.1` of some app `XYZ` installed on one device, but version `1.2` on another? If I'm reading this right, you're syncing the app install states, which I was concerned about in [#2013 (comment)](https://github.com/ImranR98/Obtainium/issues/2013#issuecomment-2528951728) - this would cause devices to show inaccurate install state info, especially for apps that do not have the "reconcile installed version..." toggle enabled. Please correct me if I'm wrong, maybe I missed something. We should be syncing the app list but not device-specific state.

If this is an issue, it would be an issue during import/export, too, as the SxncD function is just a wrapper for those calls (using whichever device has the newest settings change or app change). As I go through the above I'll take a look by making sure I have different installed apps on each device and see how it treats it, but for clarity if a change is needed in those functions I'll probably do it in a separate PR so it's easier to review independently.

ImranR98:
> If this is an issue, it would be an issue during import/export, too

Import/export is a different use case, usually infrequent and, when it happens, the user can correct mistakes manually.

With the PR as it is now, I would expect problems at least when syncing apps with the "reconcile OS version..." toggle disabled.

Feature-Requests:
How much size does this add to the app? If this is added and it makes the app noticeably bigger then an Obtainium flavor should be released for those who don't want integrations with third-party services (SxncD in this case). Or fork Obtainium and maintain a fork.

JlincFM:
> How much size does this add to the app?

Basically nothing. As mentioned in the issue ticket, all the libraries used for HTTP connection are already in the Obtainium pub list.

Feature-Requests:
We also need to trust the owners of the server.

JlincFM:
> We also need to trust the owners of the server.

It is self-hosted.

Fmstrat:
> We also need to trust the owners of the server.

It is self-hosted.

Fmstrat:
> It is an issue with import/export for apps that have the "reconcile OS version..." toggle disabled - those app entries appear with an incorrect install status and the user has to manually correct it.

Ah, I follow now, I'll investigate parsing those values on sync.

> But import/export is a different use case - usually infrequent so it's mostly fine as is. Syncing would happen all the time so we can't have that.
> 

I will say, I have a few friends using Obtainium (4 including me), and all but one of us use Import/Export for syncing between multiple phones or a tablet. So I think the use-case is happening anyway. If you want, I can look at reconciling that option period, even during import (to accompany the below)?

> My suggestion is to modify the export JSON (before upload) to remove any data that's supposed to be device-specific (mainly the installedVersion for now, maybe there are others that make sense to also exclude later). Same for import/download - merge the incoming JSON with any existing on-device data.

Good call. Will do a quick audit to see if any other options make sense, too.

ImranR98:
Yeah, modifying the original import/export function is a good idea.

mostfaxd (approved):

DeanLemans:
any news on this?

Fmstrat:
@DeanLemans Had some life stuff come up so SxncD went to the backburner for a bit. I think I'll be able to pick this up sometime in the next month.

karan:
This is clever, though I wonder if it should be possible to have the "sync" part be handled by something else (Syncthing, FolderSync etc) and have Obtanium simply do export/import on a schedule of the local file.

karan:
> @karan this can already can be done. Obtainium has "Automatically export on changes" feature in the "Import / Export" tab. The target directory can be anyone synchronized by Syncthing.


That's great! I guess manual import would be fine.

PR #2154, opened 2025-03-01
update french translation

foXaCe:
I hope the original translator won't mind that I made some modifications and corrections.

@Kevinr99089


Kevinr99089:
Not at all, I've been pretty busy lately

PR #2183, opened 2025-03-13
Add CoolApk

### LINKED ISSUE #532: CoolApk as a source
Hi, Coolapk is a Chinese app store. 
The language of site is Chinese but it has some interesting apps in this app store.
And here is another app called [UpgradeAll](https://github.com/DUpdateSystem/UpgradeAll) written in Kotlin has made it.
Would you like to take a look of the [code](https://github.com/DUpdateSystem/UpgradeAll/blob/master/core-websdk/src/main/java/net/xzos/upgradeall/core/websdk/api/client_proxy/hubs/CoolApk.kt).
Want to see it in Obtainium, so that I can manage my apps more easily.

UjuiUjuMandan:
Close #532.

Works on my machine, I'll try to workaround its geo-blocking later, so you can test it yourself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Currently supported App sources:
   - [Uptodown](https://uptodown.com/)
   - [Huawei AppGallery](https://appgallery.huawei.com/)
   - [Tencent App Store](https://sj.qq.com/)
+  - [CoolApk](https://coolapk.com/)
   - [RuStore](https://rustore.ru/)
   - Jenkins Jobs
   - [APKMirror](https://apkmirror.com/) (Track-Only)
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -319,6 +319,7 @@
     "stayOneVersionBehind": "Stay one version behind latest",
     "refreshBeforeDownload": "Refresh app details before download",
     "tencentAppStore": "Tencent App Store",
+    "coolApk": "CoolApk",
     "name": "Name",
     "smartname": "Name (Smart)",
     "sortMethod": "Sort Method",
--- a/assets/translations/zh.json
+++ b/assets/translations/zh.json
@@ -319,6 +319,7 @@
     "stayOneVersionBehind": "比最新版本晚一个版本",
     "refreshBeforeDownload": "下载前刷新应用程序详细信息",
     "tencentAppStore": "腾讯应用宝",
+    "coolApk": "酷安",
     "name": "名称",
     "smartname": "姓名（智能）",
     "sortMethod": "排序方法",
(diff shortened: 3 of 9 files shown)


UjuiUjuMandan:
User-Agent hack works; I set it to `Dalvik/2.1.0 (Linux; U; Android 9; MI 8 SE MIUI/9.5.9) (#Build; Xiaomi; MI 8 SE; PKQ1.181121.001; 9) +CoolMarket/12.4.2-2208241-universal` in Tor Browser. I wonder they won't black out their own app users when they were abroad.

It seems no special configuration is needed to test the Obtainium function.

And note that the website seems to hide certain apps; for example, a bank app https://www.coolapk.com/apk/com.icbc, it does appear in CoolApk app and Obtainium.

ImranR98:
I have no way of testing this - I get an access denied message for all APK URLs. My VPN does not have a China option, none of the other countries work.

UjuiUjuMandan:
> I get an access denied message for all APK URLs

In the app? Or in a web browser? I've tested with Tor to proxy the app, and it worked before.

UjuiUjuMandan:
I can set up a SOCKS proxy server or add a WireGuard peer on my network for your testing. Which would you prefer? Or use SNI proxies here https://github.com/ImranR98/Obtainium/issues/532#issuecomment-2721462844?

ImranR98:
I have no idea how to use the proxy since Obtainium doesn't have a proxy setting (there's an open issue iirc) and no Android-wide setting I can find either.

Either way I was able to add the app without using a proxy, it's only the APK download that fails. Code looks good though, I'm OK to merge - just remove the debug prints please.

UjuiUjuMandan:
> I have no idea how to use the proxy since Obtainium doesn't have a proxy setting (there's an open issue iirc) and no Android-wide setting I can find either.

Sorry for assuming everyone is a proxy user. The SNI proxy is utilized by manipulating DNS, replacing the original server IP with the proxy's, either by using a customized DNS server like dnsmasq or the /etc/hosts file. See https://www.bamsoftware.com/computers/sniproxy/

The latter unfortunately requires root privilege to modify on Android.

> Either way I was able to add the app without using a proxy, it's only the APK download that fails.

The APK URL eventually redirected to `imtt.dd.qq.com`, indicating they utilize Tencent App Store instead of hosting any APK on their own, at least for the ICBC bank app I've tested. This server must have different blocking rules. But Tencent's problem is that the APKs downloaded from their web interface are default 32-bit and unsuitable for my device.

> just remove the debug prints please (unless you think they're likely to be useful going forward).

Done.

Issue #1398, opened 2024-02-15
AppCenter (new source?)


pressRtowin:
Mainly interested in this due to BetterBatteryStats' release candidates being published on here (link not provided per dev's request but here's a random app I found hosted here on Google: https://install.appcenter.ms/orgs/uoft-madlab/apps/gradwell-android/distribution_groups/uoft%20general%20beta%20users), but could be useful for other apps as well. Is this possible to add as a source?

ImranR98:
All the content on that page is being loaded in with JavaScript, so Obtainium would not be able to scrape it. Do they have an API or something where we could get release info and APK links from?

cm-mc:
Would definitely be interested in this, as well; Appcenter is Plus telegram client's recommended source: https://install.appcenter.ms/users/rafalense-70ux/apps/plus-release/distribution_groups/public

nexplorer-3e:
Most of AppCenter openAPI requires a token, while there is a exception to [releases_getLatestByDistributionGroup](https://openapi.appcenter.ms/#/distribute/releases_getLatestByDistributionGroup), which most of link contains the required three args: `owner_name`, `app_name` and `distribution_group_name` (like the plusmsg and uoft-beta links mentioned above).

The apk link stores at text["download_url"], might same as `install_url`. Currently it's unclear whether the api supports multiarch.

Example usage of the API: https://install.appcenter.ms/api/v0.1/apps/uoft-madlab/gradwell-android/distribution_groups/uoft%20general%20beta%20users/releases/latest

(yes you can use latest directly mentioned by the reference)

If history package is needed, there is also a hidden API that can get "public releases" in a certain distribution group:
api/v0.1/apps/{user}/{app}/distribution_groups/{dist}/public_releases. the output should be like this:

```json
[   {
        "id": 9,
        "short_version": "1.0",
        "version": "26",
        "origin": "appcenter",
        "uploaded_at": "2022-04-19T13:40:13.181Z",
        "mandatory_update": false,
        "enabled": true,
        "is_external_build": false
    },
]
```

Example: https://install.appcenter.ms/api/v0.1/apps/uoft-madlab/gradwell-android/distribution_groups/uoft%20general%20beta%20users/public_releases

This API might be similar to https://openapi.appcenter.ms/#/distribute/releases_listByDistributionGroup, while the latter needs an token to give you "not public" releases, in theory. 

FYI how to find this api: open a public source, open devtools -> network, you would have some clues to those info.

[...]

PR #2212, opened 2025-03-30
feat(AppSource): initial AppCenter MS support

nexplorer-3e:
https://github.com/ImranR98/Obtainium/issues/1398

only implement fetching the latest one, before my code dead with M$ decision.


ImranR98:
I don't see the point of adding this given that AppCenter is apparently being retired today. If I understand correctly, this means any apps on the site will still be available but will never receive updates. So Obtainium would just be a slightly more convenient way of installing the final version of an app (no update tracking, notifications, auto updates) which is not the point of Obtainium.

PR #2232, opened 2025-04-08
Adding Hebrew

tzagim:


tzagim:
It's very interesting why you decided not to merge this pull request. Perhaps it's because Israelis are fighting against an enemy that massacred them and tried to annihilate them, and now they are succeeding and will succeed in eliminating the enemy completely!

I have removed your app from my phone and recommend every human being to do the same

PR #2269, opened 2025-04-29
Add support for vivo App Store

iKirby:
Add vivo App Store with search support. This will be useful for Chinese users.

Currently their apis used on the web (h5) version don't require any authentication. Don't know if it will change in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Currently supported App sources:
   - [Huawei AppGallery](https://appgallery.huawei.com/)
   - [Tencent App Store](https://sj.qq.com/)
   - [CoolApk](https://coolapk.com/)
-  - [vivo App Store](https://h5.appstore.vivo.com.cn/)
+  - [vivo App Store (CN)](https://h5.appstore.vivo.com.cn/)
   - [RuStore](https://rustore.ru/)
   - Jenkins Jobs
   - [APKMirror](https://apkmirror.com/) (Track-Only)
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Currently supported App sources:
   - [Huawei AppGallery](https://appgallery.huawei.com/)
   - [Tencent App Store](https://sj.qq.com/)
   - [CoolApk](https://coolapk.com/)
+  - [vivo App Store](https://h5.appstore.vivo.com.cn/)
   - [RuStore](https://rustore.ru/)
   - Jenkins Jobs
   - [APKMirror](https://apkmirror.com/) (Track-Only)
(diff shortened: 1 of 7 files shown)


ImranR98:
Will this work for other countries' versions? For example I found https://appstore.vivo.com.br
If not, the name should be changed so it's clear that this is for the Chinese store only.

Also please use the values of `hosts[0]` wherever possible instead of hardcoding host strings. See here for a good example: https://github.com/ImranR98/Obtainium/blob/main/lib/app_sources/huaweiappgallery.dart#L27
This is useful when the user decides to force the use of a specific source (source override) when the URL is not what we expect for that source. I'm guessing this is very unlikely to ever happen for the Vivo store specifically but it should still be supported for consistency.

Third minor thing - `vivo` should probably by capitalized (`Vivo`) in all the translation files.

iKirby:
Thanks for your feedback.

I didn't even know vivo has its own app store outside China, because global versions have Google Play pre-installed. I checked the BR site but I can only see subscription services, and the apis and structure is different from the Chinese site, so the support will be Chinese site only. I will add CN to the name.

I don't quite understand the use of `hosts[0]`. Vivo store's URL contains only their own `appId`, all other information needs to be retrieved from the api (package name, download URL, etc). If the URL contains a valid `appId` parameter then it is OK.

As for `vivo` or `Vivo`, their website uses `vivo` everywhere because it's their brand, so I think it should be kept the same.

And about the user agent, the default one works but I think it's better to use a browser user agent, to prevent basic bot detection. I will remove it for now, and add if required in the future.

wodin:
@iKirby if I understand correctly, using `hosts[0]` is a way to allow a user that wants to download from the Brazillian vivo store to use your vivo app source, even if yours was written to support the Chinese vivo store.

So the user would enter the Brazillian vivo store's URL into Obtainium, but then override the source to use your "vivo" app source.

I found this explanation useful:
https://deepwiki.com/search/on-a-pr-to-add-support-for-a-n_486f1496-558f-4f4a-841d-0a55b387b2e3

@ImranR98 please correct me if I am wrong.

ImranR98:
Yes that's right

iKirby:
> So the user would enter the Brazillian vivo store's URL into Obtainium, but then override the source to use your "vivo" app source.

But this is not possible. The code I added uses their api to get info, if `appId` param does not exist in the URL or it is not a valid one in the CN vivo App Store, the source will not work. It works like the [Coolapk source](https://github.com/ImranR98/Obtainium/blob/main/lib/app_sources/coolapk.dart). So using `hosts[0]` is unnecessary and it won't work.

Maybe I should use regex to match the host?

ImranR98:
Nah, you're right, it's unnecessary work. I will add a new `allowOverride: false` property so that this source (and similar ones) do not show up in the override dropdown.

PR #2290, opened 2025-05-14
allow apkpure links for apks with no variant list

### LINKED ISSUE #2147: Unable to download some apps from APKpure
**Prerequisites**
Checked

**Describe the bug**
Unable to download some apps from APKpure. For example, Doodle Jump (https://apkpure.net/doodle-jump/com.lima.doodlejump), GTA VIA (https://apkpure.net/gta-via/com.axizmotion.gtavia/download) - an error "No APK found" appears. Downloading and installing these apps manually through the website works. Other apps from APKpure can we installed as well.

**To Reproduce**
1. Try to add above apps into Obtainium.
2. See error "No APK found".

**Screenshots and Logs**

Image

**Please complete the following information:**
 - Device: Pixel 8
 - OS: Android 14 AP2A.240905.003
 - Obtainium Version: 1.1.45

**Additional context**
nv

bernikr:
this fixes #2147

### DIFF
--- a/lib/app_sources/apkpure.dart
+++ b/lib/app_sources/apkpure.dart
@@ -128,7 +128,21 @@ class APKPure extends AppSource {
               .toList() ??
           [];
       if (apkUrls.isEmpty) {
-        throw NoAPKError();
+        var link =
+            html.querySelector("a.download-start-btn")?.attributes['href'];
+        RegExp downloadLinkRegEx = RegExp(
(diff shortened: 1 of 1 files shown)


bernikr:
i just realized there seems to be a bug in downloading different architecture apks in the apkpure implementation anyway: different architectures use the same version code, but add an additional param (`nc`) to the download url, which seems to be ignored by the current code.

I'd be willing to try my hand at a rewrite of the apkpure source if that is fine with you @ImranR98

PR #2303, opened 2025-05-18
Create ar.json

Hussain96o:
Create an Arabic language file


ImranR98:
@Hussain96o what should the language menu call this? I usually use https://www.omniglot.com/language/names.htm but it shows multiple values for Arabic, I'm not sure what to pick (currently I have set it to عربى).

PR #2305, opened 2025-05-19
Rewrite APKPure Source to use the mobile API instead of web scraping

bernikr:
While developing #2290 and #2296 I discovered that the APKPure website can be pretty inconsistent when dealing with different variants, which can be both for different architectures or versions where the version number stays the same but the version code changes.

So I reverse-engineered the API used by the APKPure app, which provides a nice consistent versions endpoint, and re-implemented the APKPure Source based on that. I've tested it with all apps that were mentioned in the relevant issues and it just seems to work without any issues.

I added a new additional setting as well (`useFirstApkOfVersion`). Since some apps have many compatible apks for the same version, this setting just selects the newest of the resulting apks, allowing the user to skip the apk selection and better enable automatic updates.

I'm unsure how the localization system works, so providing the user facing string is still missing, any advice on how to do this would be appreciated.


ImranR98:
Nice, this looks really good 🎉
For translations, you need to add a new line to `assets/translations/en.json` then run `node assets/translations/standardize.js` (needs Node.js) to update the other files. I'm thinking "Auto-select first of multiple APKs" is clear - let me know if you have something else in mind or if you'd rather I update the files (will merge this then).

bernikr:
should be ready for merge now

DJCrashdummy:
have used Obtainium 1.1.53 (and noticed the bug fixed with #2296), i can report, that both 1.1.54 and 1.1.55 with this MR, work fine. :+1:
thank you very much @bernikr and @ImranR98!

PR #2436, opened 2025-07-30
Use lowercase in settings slider

shimeoki:
addresses #2418. many languages already did this.

as i checked, the settings slider for update interval is the only place where these strings are used. i updated i18n files where i was sure to change it (because of that, i skipped `de.json`).

### DIFF
--- a/assets/translations/ca.json
+++ b/assets/translations/ca.json
@@ -358,16 +358,16 @@
         "other": "{} URLs"
     },
     "minute": {
-        "one": "{} Minut",
-        "other": "{} Minuts"
+        "one": "{} minut",
+        "other": "{} minuts"
     },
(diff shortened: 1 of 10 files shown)


shimeoki:
some other things came up while writing the pr:

this slider uses `neverManualOnly` key from translation files. this sentence is written with first uppercase letter. should this be changed? i am not sure how is this for other languages, but in russian and english the letter shouldn't be uppercase after the colon.

also, `apps` key needs some attention. it also uses first uppercase letter in most translations, but is used in more than one place in the code. what's interesting, that the resulting string is almost always converted to lowercase, because of the `toLowerCase()` call. should this be changed as well? and with this change we can also remove these redundant calls in the code.

PR #2552, opened 2025-09-21
Try to handle response body as a JSON first before handling it as a p…

aminosred25:
Hello,

I am trying to setup Instagram Lite in Obtainium by directly using the following API : https://www.instagram.com/api/v1/web/lite/

The API returns a JSON and the current implementation doesn't extract the full URL since it contains unescaped characters.

The proposed change rearranges the code to try to treat the body as a JSON first before treating it as plain text.

Thank you!


ImranR98:
Yep, this makes sense

PR #2628, opened 2025-11-11
Moving russian metadata from `ru` to `ru-RU` and reformatting

Korb:
`F-Droid`, `IzzyOnDroid`, and `Google Play` work correctly with the former (in all russian-language regions), but `Google Play` doesn't work with `ru`. `ru-RU` is a universal option without any drawbacks, unlike `ru`.

Due to `F-Droid`'s way of handling line breaks, I had to merge your fully formatted with HTML tags text into a single line. An alternative is to format the list as plain text. I recommend you review the formatting of the original `full_description.txt`, as it looks terrible when viewed in the `F-Droid` mobile app. I'll leave that up to you. IzzyOnDroid handles both options perfectly, so you don't have to worry about it.

Compare the appearance of the two descriptions after the next sync with `F-Droid`, https://f-droid.org/en/packages/dev.imranr.obtainium.fdroid/ and https://f-droid.org/ru/packages/dev.imranr.obtainium.fdroid/


Korb:
> Compare the appearance of the two descriptions after the next sync with `F-Droid`,

@ImranR98:

| English | Russian |
| ------------- | ------------- |
| Screen Shot 2025-12-14 at 17 31 20 | Screen Shot 2025-12-14 at 17 31 30 |

PR #2662, opened 2025-12-05
Dismiss update notification after app install  

tnswo561412:
Summary

-Dismiss "Updates Available" notification when user installs an app from within Obtainium

Problem

When a user receives an update notification and then opens Obtainium (not via the notification) to
install the update, the notification remains even after successful installation. Steps to reproduce:

1. Receive an update notification about an application
2. Open Obtainium directly (not by tapping the notification)
3. Install the update
4. Notification is still shown

Solution

Cancel the UpdateNot if icat ion (ID: 2) after successful app installation in
downloadAndInstal ILatestApps().

Test Plan

1.Receive update notification for an app
2.Open Obtainium without tapping the notification
3.Install the app update
4.Verify notification is dismissed

Since it's difficult for me to test, I would appreciate it if you could test it.

This was written with the help of an LLM. Please bear that in mind.

### DIFF
--- a/lib/pages/apps.dart
+++ b/lib/pages/apps.dart
@@ -276,11 +276,23 @@ class AppsPageState extends State<AppsPage> {
         );
       } else if (settingsProvider.sortColumn ==
           SortColumnSettings.releaseDate) {
-        result =
-            (a.app.releaseDate)?.compareTo(
-              b.app.releaseDate ?? DateTime.fromMicrosecondsSinceEpoch(0),
-            ) ??
-            0;
+        // Handle null dates: apps with unknown release dates are grouped at the end
+        final aDate = a.app.releaseDate;
+        final bDate = b.app.releaseDate;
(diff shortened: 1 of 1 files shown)


tnswo561412:
Please review the PR.

ImranR98:
If I'm reading this right (haven't tried running it), it will remove any update notification regardless of which app(s) were updated. This can be an issue since the same notification may include multiple apps.

tnswo561412:
After reviewing the code, it seems that's the case.

tnswo561412:
Instead of bundling all app updates into a single notification, as is currently done, I propose a method where individual notifications are sent for each app that requires an update. When this method is applied, if a user updates just one specific app, only that app's notification will disappear, while the notifications for the other apps that have not yet been updated will remain. We expect this to naturally solve the problem of users overlooking the remaining updates.

PR #2668, opened 2025-12-07
Added moddroid as a source

divinezmuz:
Tested working with this build I made.

[app-arm64-v8a-normal-release.zip](https://github.com/user-attachments/files/24124198/app-arm64-v8a-normal-release.zip)

Screenshot_1
Screenshot_2

### DIFF
--- a/lib/app_sources/moddroid.dart
+++ b/lib/app_sources/moddroid.dart
@@ -1,8 +1,6 @@
 import 'package:html/parser.dart';
-import 'package:http/http.dart';
 import 'package:obtainium/custom_errors.dart';
 import 'package:obtainium/providers/source_provider.dart';
-import 'package:obtainium/app_sources/html.dart';
 
 class Moddroid extends AppSource {
   Moddroid() {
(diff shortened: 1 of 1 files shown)


ImranR98:
There's always a risk of malware when it comes to mod sites. I usually use this list to decide whether or not a site is OK to add: https://fmhy.net/mobile

Moddroid isn't there so I'm not sure about this. I can't find any info about it being malicious though so I'll approve this. Just leaving this comment here as a notice. We can always remove it if it turns out to be a bad idea. As always it's the user's responsibility to trust what they download.

D3SOX:
And here we are: Malware was distributed through this site.
Citing from https://www.bleepingcomputer.com/news/security/new-android-malware-uses-ai-to-click-on-hidden-browser-ads/

> In addition to the Xiaomi-hosted apps, the trojans are distributed via third-party APK sites (e.g., Apkmody and Moddroid, altered versions, the so-called mods, of the original Spotify, YouTube, Deezer, and Netflix apps.
> 
> The [researchers say](https://forum.drweb.com/index.php?showtopic=339666#entry923156) that most apps on Moddroid’s “Editor’s Choice” page are infected.
> 
> Infected APK files are also distributed through Telegram channels, some app examples including Spotify Pro, Spotify Plus – Official, Moddroid.com, and Apkmody Chat.

I would advise removing support for this shady site!

ModDroid is also [classified as dangerous by AdGuard](https://reports.adguard.com/en/moddroid.com/report.html?aid=16593) (using [AdGuard Home](https://github.com/AdguardTeam/AdGuardHome) here and got a warning when I tried to access the site)

+1 for FMHY, I suggest sticking to their recommended sites and not adding untrusted stuff like this in the future

ImranR98:
Thanks @D3SOX, I wouldn't have caught this. Sticking to FMHY for mod sites going forward.

D3SOX:
You're welcome. I already had a bad feeling in my gut when I saw this PR and the site initially. Only found out about this because it was mentioned in [the latest episode of Surveillance Report](https://youtu.be/2LHS-bZx3TI?t=2371) on Techlore's YT channel btw.

PR #2674, opened 2025-12-11
Fixes the bug where apps without release date info (like F-Droid apps) would appear in random positions when sorting by release date.

tnswo561412:
The previous code had an issue where comparing null dates would return 0, keeping the original order - which caused the sorting to break whenever apps were added or removed.

Now null dates are handled explicitly:
- Both null: sort by name for consistency
- One null: push the null one to the end
- Both have dates: compare normally

Apps with unknown release dates are now always grouped at the bottom (sorted by name within that group), regardless of ascending or descending order.

### DIFF
--- a/lib/pages/apps.dart
+++ b/lib/pages/apps.dart
@@ -276,11 +276,23 @@ class AppsPageState extends State<AppsPage> {
         );
       } else if (settingsProvider.sortColumn ==
           SortColumnSettings.releaseDate) {
-        result =
-            (a.app.releaseDate)?.compareTo(
-              b.app.releaseDate ?? DateTime.fromMicrosecondsSinceEpoch(0),
-            ) ??
-            0;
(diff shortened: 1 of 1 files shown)


tnswo561412:
Hello, just checking in on this PR. I understand you might be busy, but I’d appreciate it if you could take a look when you have a moment. I'm happy to resolve any conflicts or make changes if needed.

tnswo561412:
I understand you must be busy, but I would highly appreciate it if you could spare some time to review this PR. As a part of a university course assignment, feedback or a merge is crucial for my submission deadline.

Please let me know if there are any issues I need to resolve, or if you have specific feedback on the code. Thank you for your time!

ImranR98:
Hi, yeah I'm busier than usual this month. I haven't had time to actually test this PR but the code looks fine, I don't see why it wouldn't work. I'll approve and merge on that basis, things can always be corrected later if needed.

PR #2683, opened 2025-12-17
Handle existing apps for deep links

micahmo:
This PR enhances the `obtainium://add...` deep linking functionality. I found that if I followed a link for an app that I already had installed, it still prompted me to add it again. This PR improves that functionality so that, if the app that the user is trying to "add" is already installed, we navigate to the existing app page instead of the add app page.

### Before

https://github.com/user-attachments/assets/1840af35-0929-4c93-9868-ce63eb0f3720

### After

https://github.com/user-attachments/assets/bbcc1586-7fa9-4574-8eb6-2ce517af6f68

### DIFF
--- a/lib/pages/apps.dart
+++ b/lib/pages/apps.dart
@@ -1336,22 +1336,17 @@ class AppsPageState extends State<AppsPage> {
   void openAppById(String appId) {
     AppsProvider appsProvider = context.read<AppsProvider>();
 
-    AppInMemory? match;
-    for (AppInMemory a in appsProvider.apps.values) {
-      if (a.app.id == appId) {
-        match = a;
-        break;
-      }
-    }
+    AppInMemory? app = appsProvider.apps[appId];
 
-    if (match == null) {
+    // Should exist, since we just looked it up, but just in case...
+    if (app == null) {
       return;
     }
 
     Navigator.push(
       context,
       MaterialPageRoute(
-        builder: (BuildContext context) => AppPage(appId: match!.app.id),
+        builder: (BuildContext context) => AppPage(appId: app.app.id),
       ),
     );
   }


ImranR98:
The current functionality seems more useful to me since the user might want to replace their existing config with the new one (I know I've done this before). But I can see how that would be confusing if it wasn't their intention. Maybe we should have a confirmation box to let the user either overwrite their current config or not.

micahmo:
> The current functionality seems more useful to me since the user might want to replace their existing config with the new one

Ok fair enough, I hadn't considered that work flow!

How about we still navigate to "add" by default, but if the same app URL is already installed, we show a snackbar that tells the user and lets them navigate to the app page. If you think that approach would work I can try it out.

ImranR98:
Oh hold on, this is about `obtainium://add` which just takes you to the add app page, not `obtainium://app` which is what I was thinking of (that imports a whole pre-built app config). Should've paid more attention. For `add` I think your proposal makes sense as is.

micahmo:
> How about we still navigate to "add" by default, but if the same app URL is already installed, we show a snackbar that tells the user and lets them navigate to the app page.

Here's what this approach would look like, if you want to consider this instead.

https://github.com/ImranR98/Obtainium/compare/main...micahmo:feature%2Fimprove-deep-links-2

https://github.com/user-attachments/assets/980a68ea-d460-406c-94f9-a6875a84e7a5

lib/pages/apps.dart
```diff
@@ -1332,6 +1332,29 @@ class AppsPageState extends State<AppsPage> {
           : [getFilterButtonsRow()],
     );
   }
+
+  void openAppByStandardizedUrl(String standardizedUrl) {
+    AppsProvider appsProvider = context.read<AppsProvider>();
+
+    AppInMemory? match;
+    for (AppInMemory a in appsProvider.apps.values) {
```
ImranR98: This would be more performant if we could reference the app ID instead of URL, then we could do so directly and not have to loop. Which is doable since we already search through these apps in `home.dart:202`.

ImranR98 (changes requested):

ImranR98:
@micahmo just need a couple of minor changes before I merge this.

micahmo:
Hey @ImranR98, thanks for the review! I believe I addressed your comments in the way you were asking. Let me know!

ImranR98:
Thanks, works well! Merged.

PR #2686, opened 2025-12-22
Show certificate hash(es) on app page

lagodimos:


pubspec.yaml
```diff
@@ -87,6 +87,11 @@ dependencies:
   flutter_typeahead: ^5.2.0
   battery_plus: ^7.0.0
   flutter_charset_detector: ^5.0.0
+  installed_apps:
+    git:
+      url: https://github.com/sharmadhiraj/installed_apps
```
ImranR98: I would rather not rely multiple package info plugins (we already have too many plugins as it is). Could this replace `android_package_manager` (or does `android_package_manager` have support for hash checking)?

ImranR98:
Thanks @lagodimos, this works, merged.

PR #2726, opened 2026-01-18
Fix existing Signal configurations that use the website source

### LINKED ISSUE #2715: Signal.org no longer working since after 7.68.4
Pixel 9 running GrapheneOS

I have been getting updates for signal from:
https://updates.signal.org/android/latest.json

Source for that page is:

```
{"versionCode": 163601, "versionName": "7.69.2", "sha256sum": "64286cf69f100b7fdc5611d1b6a392e8457e0389355b6995af350cd452c6c59d", "url":  "https://updates.signal.org/android/Signal-Android-website-prod-universal-release-unsigned.apk", "uploadTimestamp": 1768516283395}
```

I've used version string extraction regexp, which was working until version 7.68.4 (it's now, as you see, 7.69.2): 

`(\d+\.){2}\d+`

I'm trying: 

`\d\.\d\d\.\d`

neither are working.  The error is, "Could not determine release version."  It certainly works when I search using that regexp in my editor.  The only switch I have on is "Attempt to filter APKs by CPU architecture if possible."

Can anyone make out what's happening?

eqilnou:
Closes #2715
Closes ImranR98/apps.obtainium.imranr.dev#1136

Existing configurations for the Signal app that use the source from the Signal website (`https://updates.signal.org/android/latest.json`) were broken starting from Signal version 7.69 because the apks stopped including the version name in the filenames.

I made the change in the file `source_provider.dart` by adapting the previous Signal configuration change that was there. The changes are:
- Set `"versionExtractWholePage"` to `true`; that is: toggle on the option _Apply version string extraction Regex to entire page_.
- Change the RegEx filter `"versionExtractionRegEx"` (which is the option _Version string extraction RegEx_) to the following, to extract the field `versionName` from the JSON:
  ```regex
  "versionName"\s*:\s*"([^"]*)"
  ```
- Set `"matchGroupToUse"` (which is the setting _Match group to use for version string extraction RegEx_) to `$1`.

I don't know if this kind of migration is something that needs to be done by the Obtainium app, I am just putting this here in case it could be useful.

I am not sure the changes I made are correct. Also, I put them after the previous Signal modification, instead of replacing that.

Note that configurations that get Signal from GitHub releases are unaffected by this change. Also unaffected are existing configurations that have `versionExtractWholePage` already set to `true`.

(See also ImranR98/apps.obtainium.imranr.dev#1131 for updating the configuration offered from [https://apps.obtainium.imranr.dev/](https://apps.obtainium.imranr.dev/))

### DIFF
--- a/lib/providers/source_provider.dart
+++ b/lib/providers/source_provider.dart
@@ -238,6 +238,20 @@ Map<String, dynamic> appJSONCompatibilityModifiers(Map<String, dynamic> json) {
           '\\d+.\\d+.\\d+';
       additionalSettings = replacementAdditionalSettings;
     }
+    // Signal from the website from before the apk filenames stopped including the version (#2715)
+    if (json['url'] == 'https://updates.signal.org/android/latest.json' &&
+        json['id'] == 'org.thoughtcrime.securesms' &&
+        json['overrideSource'] == null &&
+        additionalSettings['versionExtractWholePage'] == false &&
+        json['lastUpdateCheck'] != null) {
+      var replacementAdditionalSettings = getDefaultValuesFromFormItems(
+        HTML().combinedAppSpecificSettingFormItems,
+      );
+      replacementAdditionalSettings['versionExtractWholePage'] = true;
+      replacementAdditionalSettings['versionExtractionRegEx'] = '"versionName"\\s*:\\s*"([^"]*)"';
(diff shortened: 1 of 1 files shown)


ImranR98:
Hi
I don't think it makes sense to include a patch for this in Obtainium itself (the existing "migrations" are for changes to the sources within Obtainium -- Signal just uses the HTML source now, not its own dedicated source).

PR #2740, opened 2026-01-19
Translation: update hu.json

summoner001:
Fix auto translated strings.

@ImranR98 : please disable libretranslate. Much more worse than googletranslate.
It's even better if the text stays in English.

sign = jel
signers = aláírók
signature = aláírás
hash = kivonat but in this case libretranslate succesfully translated it as hamisítvány what in hungarian means the following: fake, not original.

Or use DeepL if DeepL-API free tier is avalable.


ImranR98:
Yep, not the first time I'm getting that feedback. I will disable auto-translations going forward, thanks.

PR #2748, opened 2026-01-26
i18n: Update zh-Hant-TW

xlionjuan:
B2w, submit translations with the language you don't understand is not a good ideas.


ImranR98:
Not the first time I'm getting that feedback. Will stop auto-translating strings going forward.

PR #2767, opened 2026-02-02
Add workflow to validate translation JSON

headquarter8302:
This PR adds a GitHub Actions workflow `Validate translation files`, which will execute `jq` against added and modified JSON files within the `assets/translations` directory.

The steps are:

1. Checkout repo
2. Fetch the JSON files from the PR through viewing the diff, and only those that are in `assets/translations`
    - Also send filenames to Actions output
3. Perform silent validation of JSON files from previous step's output

I've noticed that some updates to the app are *only* for fixing broken translation files, I thought this workflow should help catch these problems before they're released

### DIFF
--- a/.github/workflows/translation-validate.yaml
+++ b/.github/workflows/translation-validate.yaml
@@ -1,0 +1,31 @@
+name: Validate translation files
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6.0.2
+        with:
+          fetch-depth: 0
+
+        # we only need the JSON files
+      - name: Get JSON PR files
+        id: changed
+        # get only the changed files
+        run: |
+          echo "files<<EOF" >> "$GITHUB_OUTPUT"
+          git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -E '^assets/translations/.*\.json$' || true >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


TotallyAvailable:
Also *please* remember to drop the GITHUB_TOKEN permissions to the absolute **minimum** required (at least) whenever triggered by pull requests and/or Artifacts are uploaded. 'Secret Source' and 'persist-credentials' ***should not be left on their defaults.*** Ideally the scope should also be set to never ever allow external modifications to the workflows folder.
- for ref: ArtiPACKED, Shai Hulud and the recent Trivy breach/takeover

...how the current default Token config ever made it into production at such a "small" startup like GitHub ... absolute madness @ImranR98 

Besides some interesting "discoveries" ...(Unrelated to this app of course but not off topic) It might be worth considering adding "distributionSha256Sum=" to the gradle-wrapper.properties as well as pinning actions versions to the specific commit instead of tag and obviously never ever to ```@main``` or ```@latest``` (ref: tj-actions highjack)

The impact of adding some kind of attestation to the release workflow is unfortunately outside my expertise. Some open source apps do it besides hardened runners and other checks. In the end it comes down to the age-old topic of increasing the potential attack surface, increased maintenance and complexity vs the "confidence" gained.

Also please consider adding the (latest) file checksums to the readme (along potentially suggesting the use of something like 'DeadHash' for validation (on device)) if immutable releases do not fit into your work/release workflow. The release description as well as the (all or some) assets can obviously silently (and temporarily) be changed. Taking the date shown should never be taken as "proof" of legitimacy (which the GitHub app for example doesn't even show anyway).

I do hope those "suggestions" fit firmly enough into "good idea" instead of being seen as too much of a "demand" or chore.

PR #2906, opened 2026-04-28
Сentralize logging

dordzhiev:
Eliminate avoid_print analyzer findings and standardize logging behavior.
Preserve actionable failure context in production and background scenarios where errors were previously swallowed or only shown in UI.
Improve maintainability by routing all operational logs through a single abstraction.


ImranR98:
I would rather not add all this complexity (a whole custom logging class) for this. My guess is there are existing packages that would be a better fit if we wanted to improve logging.

dordzhiev:
> I would rather not add all this complexity (a whole custom logging class) for this. My guess is there are existing packages that would be a better fit if we wanted to improve logging.

Creating a custom class is essential for centralizing our logging logic. As for existing packages, that’s why I wrapped the external library in a Facade. Furthermore, the database interactions are now isolated in a separate class to follow the SoC principle.

ImranR98:
you're right, i closed this too quickly. I will take a closer look.

ImranR98:
Sorry about the timing, this was posted shortly before a lot of changes to the main branch. Please resolve the conflicts if possible. Otherwise I looked at the new logger dependency and have no issues with it or the general approach.

ImranR98:
Sorry for introducing a bunch of conflicts again --  just been doing a lot of refactoring.

I had an LLM try and resolve them (+ make a few changes). Results are on this branch (works fine in my testing): https://github.com/ImranR98/Obtainium/tree/reconcile/2906
Let me know if I should just patch that onto your PR or if you want to resolve the conflicts yourself.

dordzhiev:
Sorry for the delay and thanks for merging your changes into my PR. I've merged your changes.

ImranR98:
Merged, but an LLM did find some minor issues. I'm pasting a subset of those that make sense to me -- maybe something to look at for another PR if you're interested:

Risks to weigh (from code review of the actual patches)
1. AppLogger.init() is awaited, unguarded, on the startup and WorkManager critical paths (main() and callbackDispatcher) and blocks on a 7-day purge DELETE with no index on timestamp. The old design opened the DB lazily and wrapped cleanup in try/catch. A corrupt/unopenable DB would now crash app startup, or kill the background task before its own try/catch. This is the most concrete regression vs the old code.
2. Fire-and-forget db.insert with no error handling anywhere in the new files. A persistent insert failure on the main isolate hits PlatformDispatcher.onError, whose handler itself calls AppLogger.error(...) → potential error-report loop. (Same shape existed pre-PR, but now centralized in new code.)
3. Lost the localized "cleared N logs" confirmation row on log clear (the one localized log string in the codebase is now gone).
4. Minor: DB never closed; AppLogger.error() API forces string duplication at some call sites; PrettyPrinter formatting cost on errors.

PR #2938, opened 2026-05-31
docs: add troubleshooting section to CONTRIBUTING.md

Kartalops:
Add troubleshooting section covering common setup and usage issues.


ImranR98:
I think this level of docs is excessive for the main repo, maybe you could have a separate one for users who want it. In the age of LLMs I think most people would be better served by just cloning the repo and asking questions to their agent directly anyways.

PR #2951, opened 2026-06-11
feat: Add exclusive Dhizuku installer toggle

zerofrip:
## Summary

- Adds a separate **Use Dhizuku to install** toggle, mutually exclusive with the existing **Shizuku or Sui** toggle
- Wires install flow through `setInstallerMode()` so Dhizuku and Shizuku/Sui no longer auto-fallback to each other
- Migrates existing users who had Shizuku enabled but were actually using Dhizuku (`granted_owner`) to the new Dhizuku toggle
- Updates all locale strings (`useDhizuku`, split error messages)

## Plugin dependency

Requires a fork of `shizuku_apk_installer` with explicit installer mode API:

- Fork: https://github.com/zerofrip/shizuku_apk_installer (commit `1a22c7f`)
- Upstream plugin: https://github.com/re7gog/shizuku_apk_installer

The upstream plugin always prefers Dhizuku when available, which prevents true mutual exclusivity. This PR depends on the fork until equivalent API is merged upstream.

## UI behavior

| Toggle | Backend |
|--------|---------|
| Dhizuku ON | Dhizuku only (`granted_owner`) |
| Shizuku ON | Shizuku/Sui only |
| Google Play spoof | Disabled when Dhizuku is selected (Dhizuku path ignores fake install source) |

## Test plan

- [x] Dhizuku only device: Dhizuku toggle installs; Shizuku toggle shows error
- [x] Shizuku only device: inverse behavior
- [x] Both installed: toggles use intended backend exclusively
- [x] Background updates work with either privileged installer
- [x] Existing `useShizuku` users with Dhizuku auto-migrate on first launch

Made with [Cursor](https://cursor.com)


ImranR98:
Hi, I'm not sure I see the point of this -- wouldn't fallback be a good thing and offer users more flexibility? (I don't use Shizuku/Dhizuku personally and am not involved in the development of the feature, so I might be wrong). I also would rather not depend on another fork of a package (Obtainium already depends on too many packages anyways, some seemingly unmaintained).

zerofrip:
@ImranR98 
Thanks for your reply.
With the current implementation, Dhizuku doesn't start properly and always falls back to Shizuku.
Unlike Shizuku, Dhizuku runs continuously and doesn't need to be started periodically, so the fallback effectively prevents Dhizuku from being used.
Would it be okay if I opened a PR to improve the fallback behavior? That way, there would be no need to depend on the fork I modified.

ImranR98:
yes that would be preferable, thanks!

re7gog:
Hi, I merged these changes and changed shizuku behavior a little bit) (actually users will have to reinstall apps installed with shizuku because owner now pretends to be not obtainium but android shell, but according to my observations app installs more quiet this way) oops :(, btw sorry for disappearing for a long time. I hope you will able to change this PR to proper state and merge without my help

re7gog:
Of course i always recommend testing these breaking changes before rolling out this to release)

PR #2960, opened 2026-06-19
Fix the badge_obtainium.png dimensions and bolded text for consistency with other "Get it on" badges

resurrectdev1:
Changes with before and after can be viewed in this: repo https://github.com/resurrectdev1/Obtainium-Badge-Fix


ImranR98:
Thanks for the PR! I'm going to close this one though — the other badges (GitHub and F-Droid) each use their own distinct fonts rather than a consistent one, so it's expected for Obtainium's badge to follow the same approach and use its own font too.

resurrectdev1:
Hi Imran,

Thanks for taking a look! If the bold font is the issue, I'm happy to keep that part exactly as is since that's not really my concern and purely cosmetic. 

The bigger issue is the badge dimensions. Obtainium's badge is currently 564x168, while the GitHub and F-Droid badges are around 646x250. Because the aspect ratios are different, when someone sets a fixed height (like height="80") to display multiple badges together, e.g.:

<a href="https://apps.obtainium.imranr.dev/redirect.html?r=obtainium://add/https://github.com/resurrectdev1/Grove/releases/latest">
  Get it on Obtainium
</a>

<a href="https://f-droid.org/en/packages/com.resurrect.grove/">
Get it on F-Droid</a>
</div>

the Obtainium badge ends up rendering noticeably wider/larger than the others at the same height, so it looks out of place next to them.

If you're open to it, I can leave the current font and overall style untouched and just fix the dimensions/aspect ratio so it scales consistently with the other badges. Let me know.

ImranR98:
Ah, got it. Yes we should leave the font as is and change the size, thanks!

resurrectdev1:
No problem! I opened a new pr here https://github.com/ImranR98/Obtainium/pull/2965

PR #2987, opened 2026-06-25
Reserve space for app icons to prevent layout shift when opening app

deevroman:
Currently, when launching Obtanium, there's an annoying layout shift because icons don't immediately appear. This is very annoying and makes it difficult to read the app list.

To avoid this, I suggest reserving space:

<table>
<tr>
 <th> 
Before
</th>
 <th> 
After
</th>
After

<tr>
 <td>

https://github.com/user-attachments/assets/38141679-17c7-4cc1-9708-78625e8fdd92

</td>
 <td>

https://github.com/user-attachments/assets/492e1c83-ba1f-4172-8b2a-dfe5fcf13e2b

</td>
</table>

P.S. I'm not a Flutter developer, so you might want to use a different, more correct approach. For example, I didn't understand where the current icon size came from, so I'm using a constant I've determined empirically.

### DIFF
--- a/lib/pages/apps.dart
+++ b/lib/pages/apps.dart
@@ -1271,6 +1271,7 @@ class AppIconWidget extends StatefulWidget {
 }
 
 class _AppIconWidgetState extends State<AppIconWidget> {
+  static const double _appIconSize = 56;
   late final Future<void> _iconFuture;
 
   @override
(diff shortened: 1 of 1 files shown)


deevroman:
Oh, I see you've redesigned the UI in #2989 and this problem won't happen again.

PR #3048, opened 2026-07-11
Updated pt-BR translation

NIICKTCHUNS:
Updated the Brazilian Portuguese translation.

Also, there's something wrong with the language selection for Portuguese, even when I select Brazilian Portuguese the app switches back to to the Portugal Portuguese after closing and opening the app. The automatic language selection also selects the Portugal Portuguese instead of Brazilian Portuguese

Should I open an issue about this or here its fine?


ImranR98:
yes please open an issue

NIICKTCHUNS:
Done #3052

Issue #2343, opened 2025-06-08
Add setting to customize export name


schklom:
*Describe the feature*
Add a setting to customize the auto-export file name.

**Describe alternatives you've considered (if applicable)**
Manually changing the file name or using e.g. Tasker.

**Additional context**
I use the auto-export to backup Obtainium on my cloud service to make it easy to import it on a new phone. The problem is that the number of files increases without limit.

If i could keep only one file name, the backup on the cloud would be overwritten and i wouldn't end up with hundreds of files.

PR #3079, opened 2026-07-14
Add a setting for a fixed auto-export file name (#2343)

bhavit04:
Right now auto-export always writes a timestamped file name, so if the export folder syncs to a cloud service you get a new file on every change instead of one that just gets overwritten (#2343).

This adds an optional "Auto-export file name" field under the auto-export toggle. Set it and auto-exports write to `<name>-auto.json` and overwrite that each time (the existing `-auto.json` cleanup already covers it). Leave it blank and nothing changes from today's timestamped name. Manual exports aren't touched.

Tested on a device (Android 14): with a name set it keeps one file that overwrites, and blank goes back to the timestamped name.


ImranR98:
It looks like this will apply to all exports, not just automatic ones, is that right? If so we should fix that so it only uses the name in auto mode. Also there is no need to add the `-auto` suffix since it is always in auto mode when using the user's custom name.

bhavit04:
Good catch on the second one. For the first: the custom name is already gated to auto mode (the isAuto check on the file name), so manual exports keep the timestamped name and aren't affected. Dropped the `-auto` suffix like you suggested, so auto files are just `<name>.json` now. One thing that came with it: the cleanup for old auto exports matched on the `-auto.json` ending, so I also made it delete the custom-named file before writing, otherwise SAF ends up creating `name (1).json` copies instead of overwriting.

ImranR98:
Looks good, thanks. Just need to fix the new conflicts (sorry) and then this will be merged.

schklom:
Not sure if this would be hard to do, but Molly (a popular Signal fork) has a setting to keep X backups and effectively rotates them.

So I end up with X timestamped backups at any time, and I just sync the folder to my cloud without accumulating a large amount of backup files.

It's not a massive improvement so maybe not worth the extra work, but I thought it'd be worth to mention :P

Issue #3104, opened 2026-07-16
wrong "Download size" shown for apps (at least from Uptodown and Aptoide)


DJCrashdummy:
**Describe the bug**
at least for the source Uptodown a wrong download size is shown. i can't say if Uptodown reports the correct size anyhow... but if not, it is better to display something like `Download size: N/A` than a wrong size (even if it is ridiculously small and hence easy to identify as wrong).

**To Reproduce**
1. add an app for en.uptodown.com
2. open details
3. see at the bottom a reported download size of a few KB (for an app >10MB)

**Please complete the following information:**
 - Device: many different ones
 - OS: also many different ones
 - Obtainium Version: 1.6.9

**Additional context**
i'm not sure if other sources have a similar issue, but what i saw _github.com_, _apkpure.net_ and _en.aptoide.com_ seem fine.

DJCrashdummy:
as mentioned at https://github.com/ImranR98/Obtainium/pull/3119#issuecomment-5237667278 in version 1.6.11 for apps from **en.aptoide.com** there is a wrong size shown.

PR #3119, opened 2026-07-19
Fix wrong download size preview for sources like Uptodown (#3104), Minor visual tweaks

ImranR98:
- Fix wrong download size preview for sources like Uptodown (#3104)
- Minor visual tweaks

### DIFF
--- a/lib/components/category_editor.dart
+++ b/lib/components/category_editor.dart
@@ -433,35 +433,58 @@ class _CategorySelectorState extends State<CategorySelector> {
       );
     }
 
-    return Wrap(
-      alignment: widget.alignment,
-      spacing: 8,
-      runSpacing: 8,
-      children: [
-        for (final name in names)
-          Tooltip(
-            message: tr('editCategory'),
-            child: Semantics(
-              // Expose the long-press-to-edit gesture to assistive tech.
-              onLongPress: () => _edit(name),
-              child: GestureDetector(
-                onLongPress: () => _edit(name),
-                child: FilterChip(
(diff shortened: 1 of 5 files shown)


DJCrashdummy:
@ImranR98 at least #3104 (for Uptodown) seems to be fixed with release 1.6.10. :tada: 
and the GUI was also improved quite a bit with this release. :+1: 

but there is still one issue regarding these 2 things... citing from the release notes:
> Download size shown inside the update button — The estimated APK size now appears directly in the Update or Install button text (e.g., "Update · 12 MB"), and is only looked up when an update or install is actually available.

why is that? ...because then, if you want to download the APK via "Download release asset" button, you still have no idea how big it will be (if there is no update pending ATM).

IMHO as long as you have the option to download something, the information of the size (beforehand) comes in handy.
may i suggest to either always display the size in the bottom banner where it is now (no matter the local install/update state) or simply move this information up either to the "Download release asset" button or even higher up to the other information about the APK (like version, name and date) as they more or less belong together.

DJCrashdummy:
@ImranR98 thanks for fixing #3104 and also thanks for always showing the download size!

unfortunately in version 1.6.11 for apps from **en.aptoide.com** there is still a **wrong size** shown. it varies always from about 100 KB to under 500 KB but it's definitely not correct since one of the apps i've tested this with, has usually ~20 MB.

should i open a new issue for that, append it to (and reopen) #3104 or is mentioning this here, sufficient?

ImranR98:
reopened the issue

PR #3175, opened 2026-08-07
Validate file paths in archive extraction and downloads

ethicnology:
Two file paths built from source-controlled data were not validated:

1. **Tarball extraction** (`extractTarballFile`): tar entry names were used
   verbatim to build the output path. A tarball served by a source (release
   asset, direct link…) containing an entry like `../../x` would be written
   outside the extraction directory. (ZIP extraction is not affected: it
   delegates to `flutter_archive`, whose native code rejects entries
   escaping the destination — verified.)
2. **Asset downloads** (`downloadFile` with `fileNameHasExt = true`, used by
   the "download asset" feature into the shared Download folder): the file
   name comes from the source (e.g. remote link/button text) and was used
   unsanitized, allowing `../` traversal outside the destination directory.

## What this PR changes

Suggested reading order — the commits are independent:

1. **`fix: reject tar entries that escape the extraction directory`** — each
   tar entry's normalized output path must stay within the destination
   directory, otherwise extraction aborts with an error (new `en.json`
   string). Legitimate archives are unaffected.
2. **`fix: always reduce download file names to a plain basename`** — the
   basename sanitization previously applied only when `fileNameHasExt` was
   false now applies unconditionally; empty/`.`/`..` results are rejected.

- ZIP extraction (already protected by `flutter_archive`'s canonical-path
  check).
- No changes to which files are downloaded or how archives are used after
  extraction; no new settings.

### DIFF
--- a/lib/providers/apps_provider.dart
+++ b/lib/providers/apps_provider.dart
@@ -432,9 +432,12 @@ Future<File> downloadFile(
   } else if (ext == 'attachment') {
     ext = 'apk';
   }
-  fileName = fileNameHasExt
-      ? fileName
-      : fileName.split('/').last; // Ensure the fileName is a file name
+  // Never trust a source-provided fileName: always reduce it to a plain
+  // basename so it cannot escape destDir, whatever the caller passed.
+  fileName = fileName.replaceAll('\\', '/').split('/').last;
+  if (fileName.isEmpty || fileName == '.' || fileName == '..') {
+    throw ObtainiumError(tr('unexpectedError'));
+  }
   File downloadedFile = File('$destDir/$fileName.$ext');
   if (fileNameHasExt) {
     // If the user says the filename already has an ext, ignore whatever you inferred from above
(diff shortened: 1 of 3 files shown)


lib/providers/apps_provider.dart
```diff
@@ -432,9 +432,12 @@ Future<File> downloadFile(
...
     ext = 'apk';
   }
-  fileName = fileNameHasExt
-      ? fileName
-      : fileName.split('/').last; // Ensure the fileName is a file name
+  // Never trust a source-provided fileName: always reduce it to a plain
+  // basename so it cannot escape destDir, whatever the caller passed.
+  fileName = fileName.replaceAll('\\', '/').split('/').last;
+  if (fileName.isEmpty || fileName == '.' || fileName == '..') {
+    throw ObtainiumError(tr('unexpectedError'));
```
master-bob, line 439: I think this should be using a specific error message instead of `unexpectedError`. This would help simplify debugging by future devs, if necessary. Would you please update it?

PR #3179, opened 2026-08-07
Restrict WebView navigation and external link schemes

ethicnology:
1. **App detail WebView**: the embedded WebView loading the app's source
   page allowed main-frame navigation to any `http(s)` *or* `ftp(s)` URL.
   A source page (or any script on it) could navigate the in-app WebView
   anywhere — e.g. a lookalike phishing page presented inside Obtainium's
   trusted UI.
2. **Markdown links**: changelogs and "about" texts are remote-provided
   markdown. Tapping a link passed the raw `href` to an external `VIEW`
   intent with no scheme validation (`javascript:`, `intent:`, `file:`…).

## What this PR changes

Suggested reading order — the commits are independent:

1. **`fix: keep webview navigation same-origin and http(s)-only`** — the
   navigation delegate now only allows `http`/`https` navigations that stay
   on the source page's origin (scheme/host/port). Cross-origin navigations
   are opened in the external browser instead; `ftp(s)` and non-web schemes
   are blocked. Subresource loads are unaffected.
2. **`fix: only launch safe url schemes from remote markdown`** — new
   `safeExternalUrl` helper (pure, testable) that whitelists
   `http`/`https`/`mailto` and resolves scheme-less URLs against the app
   origin (preserving the existing relative-link behavior for changelogs);
   both markdown `onTapLink` handlers use it.

- JavaScript remains enabled in the WebView (source pages need it to
  render); the initial source URL load is unchanged.
- No changes to how links are rendered, only to how taps are handled.

### DIFF
--- a/lib/pages/app.dart
+++ b/lib/pages/app.dart
@@ -23,6 +23,13 @@ import 'package:webview_flutter/webview_flutter.dart';
 import 'package:provider/provider.dart';
 import 'package:url_launcher/url_launcher_string.dart';
 
+/// Whether two URIs share the same origin (scheme, host, and port — Dart
+/// normalizes default ports for http/https).
+bool _isSameOrigin(Uri a, Uri b) =>
+    a.scheme.toLowerCase() == b.scheme.toLowerCase() &&
+    a.host.toLowerCase() == b.host.toLowerCase() &&
+    a.port == b.port;
+
 class AppPage extends StatefulWidget {
   const AppPage({
     super.key,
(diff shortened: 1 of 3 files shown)


lib/pages/app.dart
```diff
@@ -23,6 +23,13 @@ import 'package:webview_flutter/webview_flutter.dart';
 import 'package:provider/provider.dart';
 import 'package:url_launcher/url_launcher_string.dart';
 
+/// Whether two URIs share the same origin (scheme, host, and port — Dart
+/// normalizes default ports for http/https).
+bool _isSameOrigin(Uri a, Uri b) =>
+    a.scheme.toLowerCase() == b.scheme.toLowerCase() &&
+    a.host.toLowerCase() == b.host.toLowerCase() &&
```
ImranR98, line 30: Would this cause problems with subdomains? For example the app's standard URL is www.apkpure.com but the external URL is just apkpure.com? Haven't actually tried this but it seems like it could be an issue. Maybe we should only compare top level domains.

ImranR98 (changes requested):

PR #3198, opened 2026-08-13
fix(forgejo): support private overridden hosts

mdalnamer:


ImranR98 (changes requested):
This would result in GitHub-specific language in the Codeberg app's options. I think Codeberg token support should be implemented separately instead of borrowing from the GitHub class.

PR #3209, opened 2026-08-19
Fix RuStore 419

skrimix:
Added RuStore "secure session" (device info + X-Client-Signature) generation, using the key extracted from RuStore apk. The API now requires this and rejects >90% of requests without it.

Used device info from Pixel 8 Pro, Android id is randomized, session is cached in memory.

### DIFF
--- a/lib/app_sources/rustore.dart
+++ b/lib/app_sources/rustore.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:math';
 import 'dart:typed_data';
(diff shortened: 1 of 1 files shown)


flexxxxer:
Thank you! Can confirm that root cause of issue and fix correct one. Same problem spotted in alternative RuStore client.

iMiKED:
Thanks for solving this annoying 419 error!

mi6e4ka:
Thank you for significant contributions to the development of third-party rustore clients :)

iMiKED:
Just implemented your fixes and can say - you're great!
They works!
Thank you again 🤝

ImranR98:
Thanks @skrimix, I will try to get this merged and released by Sunday.

PR #3217, opened 2026-08-21
Fetch AppGallery app details via API

skrimix:
Added a minimal version of API access flow from AppGallery app.

This gives us more accurate data and proper version detection.

The downsides are heavier session init (~1.5s + 140KB API response, session data cached for 24h) and more sensitivity to backend changes.

Some things to mention:
- session uses hardcoded device profile (Pixel 8 Pro running the current version of AppGallery), with randomized device id
- a new migration resets previously forced pseudo-version
- sourceRequest is modified to let string bodies through without json encoding
- added trustedApkHosts to AppSource to skip the APK origin warning for download hosts
- old dl URL approach is kept as a fallback in case the API flow breaks


ImranR98:
2 minor issues before this can be merged:
1. The session prefs key ends in `-creds`, which collides with the secure-credential routing that may end up being merged by #3181 (keys ending in -creds are treated as secrets). It should be renamed just to be safe.
2. Some user-facing errors are hardcoded English. They need to go through `tr()` with new `en.json` entries.

skrimix:
I've corrected both.

PR #3220, opened 2026-08-23
Add Samsung Galaxy Store source

BingoKingo:


ImranR98:
Just a minor issue before merging this: user-facing strings like 'Device model' and 'CSC code' should go through `tr()` with new keys in `en.json` instead of being hardcoded. Same for 'Samsung Galaxy Store API error'.

Edit: nvm, just did it to save time.

PR #3241, opened 2026-08-25
feat(rustore): add CA certificates for RuStore

### LINKED ISSUE #3170: Certificate error when updating from the RuStore
**Prerequisites**
Couldnt find this issue

**Describe the bug**
An error occurs when updating any app from the RuStore marketplace, likely due to issues with SSL certificates on Russian websites.

**To Reproduce**
Try to update any app from Rustore

**Screenshots and Logs**
Image

**Please complete the following information:**
 - Device: Pixel 7
 - OS: Android 17
 - Obtainium Version: 1.6.10

jfk9w:
* Added CA certificates for RuStore (both Harica and Mintsifry).
* Migrated `fetchNonce` to `IOClient` in order to support certificate pinning.

Fixes #3170.


ImranR98:
Im not sure we should just keep adding more certs to this feature without transparency. I think there should be a multi-select where the user can decide what certs they want to pin (and see their contents). Would you be up for adding this?

I also want users to be able to pin their own certs they add, but that's obviously not in scope for this PR.

jfk9w:
> Im not sure we should just keep adding more certs to this feature without transparency. I think there should be a multi-select where the user can decide what certs they want to pin (and see their contents).

Ironically, I think this may be less transparent (not sure if this is the correct word) for user.
What you are suggesting implies that the user knows which CA to use for each source. This requires some knowledge.
It also means that there will still be a bunch of certificates bundled with application, but every user will have to do the same setup in order to get sources working.
Maybe this needs a little more thought? I could look into it, but I can't make any time commitments, also I have zero experience with basically any Android development :)
I may be misinterpreting your idea, though, please correct me if I'm wrong.

I still have my reservations about CA certificate expiration: current schema implies that the application would need to be updated when one of CAs expires. This seems suboptimal. Maybe we should provide not the certificates themselves, but links to obtain them? But there are obvious security concerns.

> I also want users to be able to pin their own certs they add

This sounds nice, but I think it would go better with user-defined sources. I'm not in the loop for further development plans, but for a bunch of built-in sources this seems like added complexity. I may be wrong, though.

lib/app_sources/rustore.dart
flexxxxer: You dont need to change this file at all in order to add trusting Mintsifry CA for RuStore. See [source_provider.dart, line 1353-1419](https://github.com/ImranR98/Obtainium/blob/36edf4b19d00301f9babb59457f1ff7524be59d0/lib/providers/source_provider.dart#L1353-L1419)

lib/providers/source_provider.dart
flexxxxer: Why not putting that in the place [where it was designed to](https://github.com/ImranR98/Obtainium/blob/36edf4b19d00301f9babb59457f1ff7524be59d0/lib/providers/source_provider.dart#L1353-L1368)?

flexxxxer:
@ImranR98 

> Im not sure we should just keep adding more certs to this feature without transparency. I think there should be a multi-select where the user can decide what certs they want to pin (and see their contents)

You probably dont understand why certificate pinning exist, I mean reason why... Can we discuss that in #3242 ?

flexxxxer:
And @jfk9w can you rename PR? This PR isnt adding certificate pinning, it already added...

assets/ca-certs/harica-tls-rsa-root-ca-2021.crt
flexxxxer: This is Harica TLS Root 2021 type RSA. Why you didn't added Harica TLS Root 2021 type ECC? They are both used to issue certificates, without including it you left a good chance to break working certificate pinning for RuStore

flexxxxer:
Also, your PR does not fix obtaining apps from RuStore while certificate pinning disabled

ImranR98:
> What you are suggesting implies that the user knows which CA to use for each source

Right, I forgot that these certs are pinned per-source and not just trusted universally.

But anyways I think this should be closed in favour of #3244 since it integrates with the same feature used for other domains.

PR #3186, opened 2026-08-09
Added support for installing applications from root

MootComb:

PR #3277, opened 2026-09-03
feat: Root installer

inmaldrerah:
Added a root/`su`-based installer backend.

## Capabilities

- Installs APKs through `su` and `pm`
- Installs to the current user instead of user 0
  - In multiuser setups, the owner no longer need to occasionally uninstall unwanted apps introduced by Obtainium in other user profiles
  - This is actually the reason for this backend vs Shizuku
  - _Theoretically_, Shizuku can support installing to current user, but that is upstream's work, and the upstream is even archived
- Supports installing as Google Play

## Future Work

- Update translation for `installMethodRoot` and `rootNotGranted`
- Update translation for `shizukuPretendToBeGooglePlay` to include root installer

## Disclaimer

The changes in this PR are produced with the help of DeepSeek V4 Flash & OpenCode.

I have manually audited all the changes (and made some extra edits).


master-bob:
This PR is related as well #3186 .

inmaldrerah:
> This PR is related as well #3186 .

Oh, I didn't realize there is another PR for root installer... Though that one doesn't seem to support installing to current user (which is the point I started). Maybe I should contribute this there? Or is it possible to use this instead if I include the translations?

master-bob:
> > This PR is related as well #3186 .
> 
> Oh, I didn't realize there is another PR for root installer... Though that one doesn't seem to support installing to current user (which is the point I started). Maybe I should contribute this there? Or is it possible to use this instead if I include the translations?

I'm not sure. I think support of installing to current user is very beneficial, as it allows for better profile support (and thus GrapheneOS). If you do bring the translations over, please make sure to keep attribution. Imran's been quiet on this front so.... ¯\\\_(ツ)\_/¯

ImranR98:
I'm not very focused on root/shizuku features (see the "non standard API" label) and rely on contributors to handle those, so I won't be testing this out myself. Quickly skimming the code though I don't see anything to call out. I closed the other PR since it hardcodes the user and has some other small issues. @inmaldrerah have you tested this on a real device? If so, you just need to resolve the merge conflict.

inmaldrerah:
> @inmaldrerah have you tested this on a real device?

Yes. I have been daily driving auto-updates with this fork using Root installer on my phone (Restless GSI, which is based on Graphene OS source code, so I tested the install-to-current-user mechanism as well).

I just resolved the conflicts (via rebasing) and updated more translations (with the help of AI), and that's basically all what I can translate with some confidence.

If you find it good enough, feel free to initiate the workflows and merge; if there are any further suggestions or requirements, please let me know.

PR #3297, opened 2026-09-10
Update translation for deselect message in German

jgkirschbaum:
`gewählt` is in my opinion the correct translation for this in german.


ImranR98:
Ignore previous comment -- deleted

ImranR98:
But from what I'm seeing online, this would change the meaning to be more like 'select'

jgkirschbaum:
Ah, I see. Please give me some time, I will rework my pull request. In one situation "deselect" is imho correct in the other not. I'll add an option for selected.

jgkirschbaum:
I hope the PR looks good now. Thank you for this great app.

PR #3304, opened 2026-09-12
docs:README change sig hash formatting

master-bob:
Minor change to the signature hash checksum formatting; using a code block (```) allows for the copy icon to be shown in github's UI. This improves the UX as it becomes a one-click copy.

Single code (`) formatting does not have this copy feature in github markdown renderer.


master-bob:
Nice polishing. Thanks.

